PR #6, opened 2024-01-17
Landing page

ErikAmerico:
I still have to work through the details of "Ant Design".

But it looks similar to the figma so far!

Screenshot 2024-01-17 at 1 13 29 AM


src/index.css
```diff
@@ -0,0 +1,68 @@
+:root {
```
rknizzle: Is this stuff in index.css automatically generated by vite or did you add it?

index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
```
rknizzle, line 1: Good first commit! There wasn't any unnecessary or unrelated code in it
(no response necessary)

package.json
```diff
@@ -10,6 +10,7 @@
     "preview": "vite preview"
   },
   "dependencies": {
+    "antd": "^5.13.1",
```
rknizzle, line 13: Have you used antd before? Why did you decide on that one? I don't know anything about CSS frameworks so Im just curious

src/index.css
```diff
@@ -65,4 +65,4 @@ button:focus-visible {
   button {
     background-color: #f9f9f9;
```
rknizzle: Try to start just deleting stuff rather than commenting it out before you make a commit so that theres less junk left in. Its not that big of a deal because we can go in later and remove it. 

But theres also no real need to leave it in because you can just reference the previous commit if you ever need the values again. So basically if you deleted this block and then later you wanted to see what the values were again, you can see what this file looked like in the previous commit so if its already been commited in a previous commit, and then you delete it from your file, its not actually gone forever; you can reference it if you need to.

TODO for Ryan: include video example

src/App.jsx
```diff
@@ -1,6 +1,6 @@
 import "./App.css";
 import { Outlet } from "react-router-dom";
-import Header from "./components/header";
+import Header from "./components/Header";
```
rknizzle, line 3: You changed this import to capital H Header but the file is still lowercase h header so the app failed to start. Can you just make the change and then update this commit?

rknizzle (changes requested):
Cool thanks Erik! This is looking really good so far!

I made a few comments with some questions and requested a change to one of your commits. You can reply to the comments with the answer and when you're done with the change.

I'll merge the PR once those are done. 

If any of my comments dont make any sense to you, thats totally fine and you can just reply that in the comments and we can skip it for now or we can meet up and work through it.

src/index.css
rknizzle: Actually you can disregard this for now. I'll bring it up again later but I already made a lot of other comments so we can just deal with this later.

src/App.jsx
rknizzle: So you can fix this issue and update this most recent commit so that you dont need to make a new commit to fix the branch with the command `git commit --amend` which updates the latest commit with a change rather than creating a new commit.

I made a little video to show you how you would do this and Ill make some git notes at some point and include git commit --amend

Heres a link to the example video:
https://drive.google.com/file/d/1oKS1D9l5ISVsAeTnFsY3MDmGLFbXZI-b/view?usp=drive_link

rknizzle: If you've already made more commits on your machine that arent in the PR yet then don't worry about this. That makes it more complicated

PR #11, opened 2024-02-23
Login page

ErikAmerico:


src/components/Header.jsx
```diff
@@ -1,19 +1,26 @@
-import { useState } from "react";
+import { useState, useEffect } from "react";
 import { Menu } from "antd";
+import { useNavigate } from "react-router-dom";
 import "./header.css";
 
 const Header = () => {
   const [current, setCurrent] = useState("");
```
rknizzle: Maybe rename these variables to like `currentMenuItemSelected` or something. It took me a minute to figure out what `current` meant in the context of this file

src/components/Header.jsx
```diff
@@ -1,19 +1,26 @@
...
 
+  useEffect(() => {
+    if (current === "login") {
+      navigate("/login");
+      setCurrent("");
+    } else if (current === "apiDocs") {
+      console.log("TODO: Redirect to API Docs page.");
+      setCurrent("");
+    }
+  }, [current, navigate]);
```
rknizzle: This second argument to useEffect tells the function to only run when these values change right? Does navigate being in there do anything? I wouldn't think that the value of the `navigate` variable would ever change

src/pages/Login.jsx
```diff
@@ -0,0 +1,79 @@
...
+      span: 16,
+    }}
+    style={{
+      maxWidth: 600,
+    }}
+    initialValues={{
+      remember: true,
+    }}
+    onFinish={onFinish}
+    onFinishFailed={onFinishFailed}
```
rknizzle: Just taking notes for myself here.
How does onFinishFailed know when it failed? I'm guessing if the function that runs when you submit throws an error or something? 

My guess is that in one of the next commits, you'll add a function to run when you submit and onFinishFailed will probably run if that throws an error

src/pages/Login.jsx
```diff
@@ -0,0 +1,79 @@
...
+    autoComplete="off"
+  >
+    <Form.Item
+      label="Username"
+      name="username"
+      rules={[
+        {
+          required: true,
```
rknizzle: How are the `required` values enforced? Will it bring up a pop up letting you know that a field was required if you submit with no email or password? I"ll test this out the in app to see

src/main.jsx
```diff
@@ -4,14 +4,14 @@ import App from "./App.jsx";
 import { BrowserRouter as Router, Routes, Route } from "react-router-dom";
 
 import Landing from "./pages/LandingPage.jsx";
-import Login from "./pages/Login.jsx";
+import LoginRegister from "./pages/LoginRegister.jsx";
```
rknizzle: I'll probably at some point switch to a separate /login and /register page but this is good for now

rknizzle (changes requested):
Heres a partial review. I need to finish going through the rest of the commits but Im taking a break so I'll submit what I have so far.

NOTE to self: Pick back up at commit aefe9ce055128044433eb94c799edcfa97a918d0

src/components/Header.jsx
ErikAmerico: In a later commit, I removed these variables

src/components/Header.jsx
ErikAmerico: Yes I believe you are right. I was running into an issue where after logging in, and navigating away from the login page. The "login" button in the header wouldn't change to "logout" immediately. I tried to use navigate as a dependency of the useEffect hook, thinking that when navigate was called, it would run. It actually worked, but after a little research, It wasn't due to navigate being listed in the dependency array. Maybe it caused some reacty side effect that i am unaware of. But regardless, This code was also removed in a later commit

src/pages/Login.jsx
ErikAmerico: I think required is a rule provided by and enforced by ant design. Yeah if you try and submit the form without an email or password, the form won't submit and a message will pop up below whichever field you left blank

src/components/Footer.jsx
```diff
@@ -1,9 +1,11 @@
 import { Row, Col } from "antd";
 import "./footer.css";
+import PropTypes from "prop-types";
 
-const Footer = () => {
+const Footer = ({ sticky }) => {
+  const footerClass = `footer-container ${sticky ? "sticky" : ""}`;
```
rknizzle, line 6: Why does this sometimes need to be sticky and sometimes not?

src/components/footer.css
```diff
@@ -51,3 +51,12 @@ a:hover {
     transform: scale(1.05);
   }
 }
+
+@media screen and (min-width: 992px) {
+  .footer-container.sticky {
```
rknizzle: Do spaces turn to . in the css file? I noticed that the class name in your jsx file was `footer-container sticky`

src/components/Register.jsx
```diff
@@ -1,12 +1,49 @@
...
+        body: JSON.stringify({
+          email: values.email,
+          password: values.password,
+        }),
+      });
+
+      if (!loginResponse.ok) throw new Error("Login failed after signup");
+
+      const loginData = await loginResponse.json();
+      console.log("Login Success:", loginData);
```
rknizzle: I think the HTTP request code for each request could go into their own functions. Especially the login one so that it can be reused in both the login and register form

src/components/Header.jsx
```diff
@@ -7,8 +7,49 @@ const Header = () => {
...
+  const [isLoggedIn, setIsLoggedIn] = useState(false);
+  console.log("isLoggedIn", isLoggedIn);
+
+  const confirmLogin = async () => {
+    try {
+      // Include credentials to send cookies with the request
+      const response = await fetch("/auth/confirm", {
+        credentials: "include",
+        headers: {
+          "Cache-Control": "no-cache",
```
rknizzle: What does the no-cache do?

src/components/Login.jsx
```diff
@@ -104,6 +107,7 @@ const Login = ({ toggleForm }) => {
 
 Login.propTypes = {
   toggleForm: PropTypes.func.isRequired,
+  confirmLogin: PropTypes.func,
```
rknizzle: I dont think confirmLogin is a prop, is it?

utils/authContext.jsx
```diff
@@ -0,0 +1,40 @@
+import { createContext, useState } from "react";
+import PropTypes from "prop-types";
+
+export const AuthContext = createContext();
+
+export const AuthProvider = ({ children }) => {
```
rknizzle: What's children here?

src/components/Login.jsx
```diff
@@ -23,7 +25,8 @@ const Login = ({ toggleForm }) => {
 
       const data = await response.json();
       console.log("Login Success:", data);
-      navigate("/");
+      confirmLogin();
```
rknizzle: No need to make the HTTP request to confirm login after logging in or signing up. If the login request responds with a 200 then thats enough confirmation. The confirm login request is just for checking if a user is logged in from a previous session. So I think you can just setIsLoggedIn here instead

src/components/Header.jsx
```diff
@@ -1,69 +1,38 @@
...
         },
       });
-      setIsLoggedIn(false);
       navigate("/");
-    }
-  };
-
-  useEffect(() => {
-    if (current === "loginregister") {
+      confirmLogin(); // Check login status again after logout
```
rknizzle: Same thing here, you dont need to send the HTTP request to confirm login after logging out you can just set the isLoggedIn state

src/components/Header.jsx
```diff
@@ -1,69 +1,38 @@
...
-  };
+  const { confirmLogin, isLoggedIn } = useAuth();
 
   useEffect(() => {
     confirmLogin();
-  }, [navigate, setIsLoggedIn, setCurrent]);
-
-  const onClick = (e) => {
-    setCurrent(e.key);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
rknizzle: Whats this eslint comment?

rknizzle (changes requested):
Sick these pages look really good! I mostly just have a bunch of questions because I'm trying to learn how some of this frontend stuff works but I also have a couple small change requests in the comments. 

Some other ideas that I have
- Move the utils directory into src/
- Add requirements to the password input on the register page so that it has to be 8 characters. It should just like say 'minimum 8 characters' under or next to password and when you press submit it should raise an error popup that says the password needs to be 8 characters

Also if you can figure it out; require a capital letter and a special character in the password but this isnt as important as the 8 characters

- Change the feedback on the email input field so that it doesn't say 'The input is not valid Email' right when you start typing. I think it would be better if it just said that after you press submit. And the login email text input doesnt need the valid email check at all. Only the sign up does. And I think the same idea goes for the matching password check and the missing fields check. I think it might be better if it just said that on submit rather than right as you start typing. (This isnt super important so feel free to just make a github issue for this and just do it later if you think its going to take a long time to change)
- I think it might make more sense to merge the components and pages into one components directory. It makes it harder to keep all the related code grouped next to each other when they're split between the pages and components folders. Don't do this for this PR but lets meet up or zoom this week and work on this. I made this into an issue here: https://github.com/rknizzle/propermesh-view/issues/12


Thanks dude! Let me know if you have any questions on anything or if any of these ideas don't have enough info

src/components/Footer.jsx
ErikAmerico: If a page doesn't have enough content to fill the view height, then the footer kind of tucks up under the last element  as you can see here: 

non-sticky-footer

By adding the sticky class, it sticks it to the bottom of the page: 

sticky-footer

We would keep the sticky class on the login/register page, but eventually remove it from the app page when its completed. 

The landing page has enough content to push the footer to the bottom naturally

src/components/footer.css
ErikAmerico: When i used `footer-container.sticky`, I was unnecessarily targeting any element that contained both classes `footer-container` and `sticky`.  The . is like concatenating the two classes.  It's definitely a more precise way of targeting an element, but considering there is only one situation where `sticky` would be applied, I could have left out `.footer-container` and just used the `.sticky` class to identify the target. 

PS... any or all of what i just said could be wrong. But i believe its correct

src/components/Header.jsx
ErikAmerico: When i was initially testing out the /confirm route, I would get a 304 'not modified' status code in response after /confirm was first checked due to caching. I wanted the actual 200 status code. This prevented that issue but i'm testing it out again, and am not sure if this was the solution or reason that everything started working. 

When i remove  `"Cache-Control": "no-cache",` now, I still get the 200 status code. So i may have been doing something wrong before

src/components/Login.jsx
ErikAmerico: Correct, It's not a prop. I think i may have accidentally left that in there during my trial and error phase of passing confirmLogin around.

utils/authContext.jsx
ErikAmerico: I wrapped AuthProvider around the entire application: 

children

so its children are Header, Outlet(all the pages and components) and footer (probably unnecessary at the moment, but in the future who knows). All of the children have access to the authentication context provided by AuthProvider

ErikAmerico: This is easier than passing down some sort of authentication prop from page to page. Now the entire app knows if someone is logged in or not right of the rip. 

hopefully

src/components/Register.jsx
ErikAmerico: Okay cool! Would i just create that function in its own separate file and import it into the login component as well as the register component?

src/components/Login.jsx
ErikAmerico: I can do that! I figured confirmLogin would setIsLoggedIn across the board. But I can see how there are unnecessary requests being made to the backend.

src/components/Header.jsx
ErikAmerico: This isn't really a comment, It's disabling an eslint warning. The warning wanted me to use confrimLogin in the dependency array of the useEffect hook. But I only wanted that hook to run once when the header first renders. Which is what the empty dependency array does.

A lot of this will be changed when i replace `confirmLogin` with `setIsLoggedIn` in the login and register components

src/components/Footer.jsx
rknizzle: Gotcha thanks for the screenshot! Does it cause a problem when the footer is made sticky on the pages that naturally push it to the bottom anyway? If not we could just make it always sticky so that we can get rid of this option

src/components/Register.jsx
rknizzle: Yea you can just put them into their own files like login.js or loginRequest.js and register.js or registerRequest.js or whatever you think makes sense. And then just import it in

src/components/Login.jsx
rknizzle: Ok cool. I think there are also some other files where you put things in propTypes that weren't actually props

rknizzle: You can delete them in a new commit

src/components/Header.jsx
rknizzle: Yea I'm actually not totally sure about this. I'll look into cache-control and get back to you here

utils/authContext.jsx
rknizzle: Gotcha ya so this is kind of a way to add some global state that all the components can access. Cool!

src/components/Header.jsx
rknizzle: Ok ya gotcha I see

src/components/Header.jsx
rknizzle: I still don't totally understand this but I don't think that its doing anything currently because my server doesn't do anything with the values. I could be wrong though. But lets remove it for now and see if its still working

src/components/footer.css
rknizzle: Oh ok I see so when its space separated in the HTML element, you're setting that element to both of those classes and this is how you specifically modify elements that have both of the classes. I gotcha

src/components/Header.jsx
rknizzle: Ok. I'm also guessing that you may have been getting an ESlint warning when you didn't have it based on that other commit you wrote. I wonder if thats why you added it

src/components/Footer.jsx
ErikAmerico: Yeah it causes the footer to always be visible on the screen. We could do that if its something you'd prefer, I'm personally against it. 

I'll show you the difference when we get together and you can make the final call!

rknizzle: I think another option could be to only have the footer on the landing page and not on any other pages. What do you think about that?

src/components/Register.jsx
```diff
@@ -67,36 +104,13 @@ const Register = ({ toggleForm }) => {
...
-          className="form-layout"
-          rules={[
-            {
-              required: true,
-              message: "Please input your password!",
-            },
-          ]}
-          hasFeedback
-        >
+        <Form.Item name="password" className="form-layout" hasFeedback>
```
rknizzle: Nice the error feedback shows up now after you press submit! The only thing still lingering is that theres a green check mark next to the password even when the length of the password is too short. I think we should just remove the green check mark. I think that is this hasFeedback property here?

src/components/Register.jsx
```diff
@@ -53,6 +62,43 @@ const Register = ({ toggleForm }) => {
...
+            <ul style={{ margin: 0 }}>{missingPasswordRequirements}</ul>
+          </>
+        ),
+        icon: <ExclamationCircleTwoTone twoToneColor="#eb2f96" />,
+        placement: "top",
+        duration: 15,
+        style: { width: 350 },
+      });
+      return;
+    }
```
rknizzle: Cool this feedback is really good! It looks pretty duplicated in the output though. You can just remove the top one or something
duplicate-message

src/components/Register.jsx
```diff
@@ -53,6 +62,43 @@ const Register = ({ toggleForm }) => {
       return;
     }
 
+    const hasUpperCase = /[A-Z]/.test(values.password);
```
rknizzle: Can you also think of a good place in the UI to specify that a special character and capital letter are required without needing to press submit?

src/components/landing/LandingPage.jsx
```diff
@@ -40,8 +40,9 @@ export default function Landing() {
         >
           <h2>Geometry Analysis</h2>
           <p>
-            Geometry Analysis is specifically designed to prepare your 3D models for 3D printing or
-            manufacturing by providing a comprehensive examination of their structure and integrity.
+            Geometry Analysis is specifically designed to prepare your 3D models
+            for 3D printing or manufacturing by providing a comprehensive
+            examination of their structure and integrity.
```
rknizzle, line 45: It looks like in commit 6372d26a269c963a21fb6d5b4d6c94f137ae9e84 these lines got auto-formatted to have a different number of words on each line which is fine. The different lines doesnt affect how it shows up in the browser does it?

src/components/header/Header.jsx
```diff
@@ -61,7 +61,9 @@ const Header = () => {
 
   return (
     <div className="header-container">
-      <div className="header-title">Propermesh</div>
+      <div className="header-title" onClick={() => navigate("/")}>
+        Propermesh
+      </div>
```
rknizzle, line 66: This is great

src/loginRequest.js
```diff
@@ -0,0 +1,24 @@
...
+    if (!response.ok) {
+      throw new Error("Login failed");
+    }
+
+    const data = await response.json();
+    return { loginSuccess: true, data };
+  } catch (error) {
+    console.error("Login Error:", error);
+    return { loginSuccess: false, error: error.message };
+  }
```
rknizzle: So you're using a concept here where you're returning a boolean value for if it succeeded along with the return value and thats totally valid. The programming languages Go and Rust do something similar where any function that can have an error will return both an error and a value

Here is an example of how errors are handled in Go. Functions can return both a value and an error and only one of them should actually hold some info or data.
```go
value, error = someFunctionThatSometimesErrors()
if (error != none) {
  // handle the error here
}

// otherwise this means there was no error so you can assume there is a value
doThing(value)

```

So you could also do this in javascript where you can modify every function that can return an error so that it creates an object that contains the success boolean and has an actual return value like you did in this file. 
But generally because of how javascript is designed to have errors that can be thrown, pretty much everybody just uses throwing errors as the equivalent of returning loginSuccess: false. So I think we should stick with the convention of throwing errors in our project as well. 

So basically you can just let this function throw an error if something goes wrong and then in your component, just use try catch instead of `if (loginSuccess) { } else {}`

A function should return one particular thing and if it cant do that for whatever reason then it should throw an error.


Let me know if any of this doesn't make sense or if you want more examples!

rknizzle (changes requested):
Good stuff! This is basically done and I'll deploy it to propermesh.com once its merged in.

My one other piece of feedback would be to move loginRequest.js and registerRequest.js into components/loginRegister so that all of this related code is grouped together nicely and not just in the big general src/ directory

src/components/Register.jsx
ErikAmerico: yup!! easy fix. good call

src/components/landing/LandingPage.jsx
ErikAmerico: I don't think so! I noticed that too and checked it out a little while back. We are good

ErikAmerico:
Hey, check these latest commits out!

src/components/Register.jsx
rknizzle: Ya I like the UI you went with

rknizzle:
Everything looks good. Merging it in!

PR #17, opened 2024-03-02
Little things post deploy

ErikAmerico:
Simple enough fixes!

Do you think the main logo on the landing page could be larger when viewing from mobile devices?


src/components/header/Header.jsx
```diff
@@ -27,7 +27,7 @@ const Header = () => {
     } else if (e.key === "login") {
       navigate("/login");
     } else if (e.key === "apiDocs") {
-      window.location.href = '/docs';
```
rknizzle, line 30: We'll have to at some point sync up our auto-formatters so that we don't have a bunch of little things like this leaking into our commits. I'm guessing you just use the default settings so I'll just need to set it up on my machine sometime. For now IDK if there is any way for you to just run the auto-formatter on the whole codebase at once. But if it is, before you start working on something you can just run that and commit all the changes that it makes as one commit with a title like 'Run auto formatting on codebase'

If thats no possible no worries; ill get my auto formatter setup sometime soon

src/components/landing/LandingPage.jsx
```diff
@@ -20,12 +20,6 @@ export default function Landing() {
         Software tools for automating 3D printability checks of 3D model files
       </h2>
 
-      <img
-        id="coming-soon-image"
-        src="assets/coming-soon.png"
-        alt="Coming Soon"
-      />
-
```
rknizzle: When this gets removed it puts the two other things that were above and below it next to each other rather than on top of each other. Can we format it back to on top of each other?
next-to-each-ther

rknizzle (changes requested):
Thank you! Looks good, I just have one change to the formatting of the stuff above and below the coming soon image

Oh also is it possible to make the logo and title in the header one clickable unit? Currently its possible to put your mouse in between them and it wont be clickable in that spot

Sort of like on https://www.hubs.com/ how the whole area of the logo/title are clickable and not just specifically the logo and the title separately

rknizzle:
@ErikAmerico I actually kind of like the size of the logo on mobile. I'm good with it staying like that for now

ErikAmerico:
My latest commit of *redesign* (not "resign") landing page, is not responsive yet! Just incase you check it out and it is wonky. 

It's just not done yet.

The image is just a placeholder as is everything else

ErikAmerico:
I'm still not sold on the "pre-footer" set up. 

But the website as a whole is much easier on the eyes, incase someone happened to wander onto propermesh.com. 

The commented out css (landingPage.css) shown below centers the text of each paragraph in the pre-footer instead of having each line aligned to the left of their columns.
I wasn't sure how much or what you were planning to write here so you can make the call!

image

src/components/header/Header.jsx
```diff
@@ -63,14 +63,14 @@ const Header = () => {
...
-        onClick={() => navigate("/")}
-      />
-      <div className="header-title" onClick={() => navigate("/")}>
-        Propermesh
+      <div className="header-logo-title" onClick={() => navigate("/")}>
+        <img
+          src="/assets/favicon.png"
+          alt="logo-in-header"
+          id="header-logo"
+          onClick={() => navigate("/")}
```
rknizzle, line 71: Im guessing this onClick is redundant now? Or is it doing something?

rknizzle (approved):
This page is looking really good now

src/components/header/Header.jsx
rknizzle: I think that it is based on a quick test so I'll just remove it

rknizzle:
Going to merge this in and then I'll fill in the spots with content in a separate PR and send it over to you!

PR #22, opened 2024-03-07
Add content to redesign

rknizzle:
Initial idea of the image in the prefooter rather than a second paragraph. I still need to fill in the text though

How do you feel about the prefooter with the image in it?

Id also like to add a title to the paragraph in the prefooter I think here: Something like 'A Powerful API'
title-to-prefooter

Also the picture would probably look better more to the right

Also I think the paragraph would look better if the font was slighty larger and the spacing between lines is a little bigger so it takes up more space. And keeping it in more of a square shape.
Sort of like this:

example


ErikAmerico:
Nice!

I like the image idea.

I'll get back in there tonight and make some final adjustments.

ErikAmerico:
This auto-formatted again. I'll figure out how to run the formatting on the entire project automatically like you talked about before

ErikAmerico:
@rknizzle I think we're good for now! check it out.

rknizzle:
Awesome! I don't have any feedback on the code because I'm not great with css but I think it looks really good!

I'll update some of the information in it over time but I think its really good for now so I'll deploy this new version today

PR #24, opened 2024-03-12
Upload part button

ErikAmerico:


rknizzle (approved):
Nice good job! 

Lets put all of these components together into one directory though so they are all grouped up. 

something Like:
```
src/components/AnalysisPage/
    AnalysisPage.jsx
    UploadPartButton.jsx
    ViewPartsButton.jsx
```

I'm just going to merge this PR in right now though so you can just make this change whenever in one of your other branches

PR #25, opened 2024-03-12
View parts button

ErikAmerico:


src/components/viewPartsButton.jsx
```diff
@@ -16,23 +16,69 @@ const ViewParts = () => {
...
+      .then((res) => res.blob())
+      .then((blob) => {
+        const a = document.createElement("a");
+        const href = URL.createObjectURL(blob);
+        a.setAttribute("download", fileName);
+        document.body.appendChild(a); // apparently, this line is needed to make 'a.click()' work in firefox/ensures that the link is part of the document
+        a.href = href;
+        a.click(); // Trigger the download
+        document.body.removeChild(a); // Clean up by removing the link from the document
+        URL.revokeObjectURL(href); // Free up memory by revoking the object URL
```
rknizzle, line 30: These lines are kind of complicated to read through so they might be good to extract into a new function like downloadBlobToLocalMachine or something and put all the lines into that so that it reads easier in here

rknizzle (approved):
Works very good so far!

I left a quick review comment but Im just going to merge this in now. You can extract that code into a function in a later commit if you think that makes sense

PR #26, opened 2024-03-13
3D model viewer

ErikAmerico:
I wasn't sure how to leave a comment on a specific commit without creating a pull request.

This was a big commit because it was mostly trial and error that eventually worked. Its a starting point

The 3d model viewer is has a ton of react elements at the moment, But i wanted to just get it working and then I'll try and refactor it to be more vanilla.

But this is kinda cool so far! I've only uploaded STL files but they are working. 

There is still a long way to go even without taking into account scaling back the reacty parts.


src/components/3D-model-viewer.jsx
```diff
@@ -0,0 +1,93 @@
...
+
+      const controls = new OrbitControls(camera, renderer.domElement);
+      controls.addEventListener("change", () => renderer.render(scene, camera));
+
+      const animate = () => {
+        requestAnimationFrame(animate);
+        renderer.render(scene, camera);
+      };
+      animate();
+    });
```
rknizzle: I think the only Reacty things in this logic is that its inside the useEffect, it grabs the canvasRef from the Component state, and it gets the objectUrl from state as well. So I think a good starting point for making this not Reacty is to just extract this logic into a function and just pass in the canvas and the objectUrl as arguments and then just call the function inside the useEffect. I could be missing something though so its possible its more complicated than that

src/utils/3dModelFileContext.jsx
```diff
@@ -0,0 +1,18 @@
...
+  return (
+    <ModelFileContext.Provider value={{ fileFor3dModel, setFileFor3dModel }}>
+      {children}
+    </ModelFileContext.Provider>
+  );
+};
+
+ModelFileProvider.propTypes = {
+  children: PropTypes.node,
+};
```
rknizzle: I think that using Context for the file data is a bit overkill. From what I understand, context is usually used for things that you want to be able to be used globally by all of your components but the file data is only really used by stuff in the AnalysisPage so I think the variable holding the file data would be better as state in the AnalysisPage component and then it can be passed into the model viewer component as a prop. And then I think the setter function would be passed into the upload button as a prop too. I'm not too experienced with this stuff though so let me know if theres a reason why you prefer using a context in this case.

Also you can ask me any questions here or in a zoom meeting if you want more details about state/props vs context

rknizzle (changes requested):
Sick! Good job getting this working already! The analysis page is coming together pretty fast

I'll try to start speccing out more of the AnalysisPage tonight since you're moving through these.

And I think here are some examples of using different types of loaders with threejs:
https://github.com/mrdoob/three.js/tree/4639edd29facefbe5fa3aff99d105c6a9066b895/examples/jsm/loaders

I think the GLTF loader might also do the GLB files. Don't worry about getting all of these working though. If you run into any problems with any of the file types just let me know

Let me know if you have any other questions or aren't sure what to do next

src/utils/3dModelFileContext.jsx
ErikAmerico: Yeah you are absolutely correct, I think it may have just gotten late and i remembered from the bootcamp that you're not supposed to pass props upwards. I confused myself and just ended up creating a context file to take care of this.

I updated the state to be defined in the analysis page file!

src/components/viewPartsButton.jsx
```diff
@@ -1,10 +1,23 @@
...
+  //not sure if this fetch call was the best approach
+  //It's the only way i could get the entire file at the moment
+  const uploadPartTo3dModel = (part_id) => {
+    fetch(`/api/v0/parts/${part_id}/file`)
+      .then((res) => res.blob())
+      .then((blob) => {
+        const file = new File([blob], fileNameForUpload, { type: blob.type });
+        setFileFor3dModel(file);
+      });
+  };
```
rknizzle: Yea nice this seems to be the correct approach to me

rknizzle: I think this function name could be improved though. Because its actually downloading a file from the server so starting it with upload is kind of confusing. I think what you're meaning by upload is that its moving the file data into the 3d viewer component. So maybe call it downloadPartFileAndPlaceIn3dViewer or something like that.

src/components/analysis/viewParts/viewPartsButton.jsx
```diff
@@ -33,14 +34,7 @@ const ViewParts = ({ setFileFor3dModel }) => {
     fetch(`/api/v0/parts/${part_id}/file`)
       .then((res) => res.blob())
       .then((blob) => {
```
rknizzle: I think that changing the name of downloadFile -> downloadFileToLocalMachine or something could be useful as well to differentiate it from the other function that downloads the file into a blob and uses it for the viewer

rknizzle (commented):
Just a couple notes here on the new commits but nothing big. Codes still looking really good!

ErikAmerico:
@rknizzle 

After moving the threeJS set up to its own function, I was getting the error 
`[.WebGL-0x12408688500] GL_INVALID_OPERATION: Vertex buffer is not big enough for the draw call` When trying to view a 2nd model in the viewer.

Which from what i could understand pretty much meant that there was left over data from the previous model. The useEffect hook has a built in cleanUp function that takes care of this issue for us. 

Here is a detailed explanation of whats going on provided by the robot: 

>In React, a cleanup function is a mechanism provided by the useEffect hook to perform cleanup operations when a component unmounts or before the effect runs again due to changes in its dependencies. This is crucial for avoiding memory leaks, removing event listeners, canceling API calls, or any other cleanup necessary to prevent unintended behavior in your application.

>How Cleanup Functions Work:
Provided by useEffect: Within the useEffect hook, you can return a function at the end of the effect. This returned function is the cleanup function. React will store this function and call it at the appropriate time for cleanup.

>When are they called:

>Before re-running the effect: If the component remains mounted and the effect's dependencies change, React will call the cleanup function before re-invoking the effect function to ensure that the component cleans up after its previous render before setting up new side effects.
On component unmount: If the component is going to unmount from the DOM, React calls the cleanup function to perform any necessary cleanup operations, such as removing event listeners or canceling subscriptions and timers, to avoid memory leaks.

>Example Explanation:

>In the provided code, there are two useEffect hooks used within the ModelViewer component. Each has its own cleanup mechanism:

>Cleanup for Object URL:

>When a file for a 3D model is provided, an object URL is created using URL.createObjectURL to be used as a source for the 3D model viewer. This URL represents a temporary object in the browser's memory.
The cleanup function here calls URL.revokeObjectURL(url) when the component unmounts or the fileFor3dModel changes. This is necessary to release the memory and resources allocated for the temporary object URL, preventing memory leaks.

>Cleanup for Three.js Scene:

>The second useEffect sets up a Three.js scene when objectURL is available. It initializes a 3D scene, camera, lights, and model loader. The setupThreeScene function returns a cleanup function.
This cleanup function iterates over disposables (like geometries, materials, meshes, and controls) to dispose of them properly, calls renderer.dispose() to clean up internal Three.js renderer resources, and clears the scene. This cleanup is crucial for freeing up WebGL resources and preventing memory leaks, especially when dealing with complex 3D graphics and frequent component updates or unmounting.

>Summary:

>The cleanup function in React's useEffect hook is a powerful feature for managing and cleaning up side effects in functional components. It ensures that your application remains efficient, leak-free, and that resources are properly managed across component lifecycles.

rknizzle:
> @rknizzle
> 
> After moving the threeJS set up to its own function, I was getting the error `[.WebGL-0x12408688500] GL_INVALID_OPERATION: Vertex buffer is not big enough for the draw call` When trying to view a 2nd model in the viewer.
> 
> Which from what i could understand pretty much meant that there was left over data from the previous model. The useEffect hook has a built in cleanUp function that takes care of this issue for us.
> 
> Here is a detailed explanation of whats going on provided by the robot:
> 
> > In React, a cleanup function is a mechanism provided by the useEffect hook to perform cleanup operations when a component unmounts or before the effect runs again due to changes in its dependencies. This is crucial for avoiding memory leaks, removing event listeners, canceling API calls, or any other cleanup necessary to prevent unintended behavior in your application.
> 
> > How Cleanup Functions Work:
> > Provided by useEffect: Within the useEffect hook, you can return a function at the end of the effect. This returned function is the cleanup function. React will store this function and call it at the appropriate time for cleanup.
> 
> > When are they called:
> 
> > Before re-running the effect: If the component remains mounted and the effect's dependencies change, React will call the cleanup function before re-invoking the effect function to ensure that the component cleans up after its previous render before setting up new side effects.
> > On component unmount: If the component is going to unmount from the DOM, React calls the cleanup function to perform any necessary cleanup operations, such as removing event listeners or canceling subscriptions and timers, to avoid memory leaks.
> 
> > Example Explanation:
> 
> > In the provided code, there are two useEffect hooks used within the ModelViewer component. Each has its own cleanup mechanism:
> 
> > Cleanup for Object URL:
> 
> > When a file for a 3D model is provided, an object URL is created using URL.createObjectURL to be used as a source for the 3D model viewer. This URL represents a temporary object in the browser's memory.
> > The cleanup function here calls URL.revokeObjectURL(url) when the component unmounts or the fileFor3dModel changes. This is necessary to release the memory and resources allocated for the temporary object URL, preventing memory leaks.
> 
> > Cleanup for Three.js Scene:
> 
> > The second useEffect sets up a Three.js scene when objectURL is available. It initializes a 3D scene, camera, lights, and model loader. The setupThreeScene function returns a cleanup function.
> > This cleanup function iterates over disposables (like geometries, materials, meshes, and controls) to dispose of them properly, calls renderer.dispose() to clean up internal Three.js renderer resources, and clears the scene. This cleanup is crucial for freeing up WebGL resources and preventing memory leaks, especially when dealing with complex 3D graphics and frequent component updates or unmounting.
> 
> > Summary:
> 
> > The cleanup function in React's useEffect hook is a powerful feature for managing and cleaning up side effects in functional components. It ensures that your application remains efficient, leak-free, and that resources are properly managed across component lifecycles.

@ErikAmerico Is that still an issue you're having or did you solve it?

ErikAmerico:
@rknizzle 

All is good now!!

I didn't fully understand the cleanup situation at first, that's why I copied that explanation in there in case you also didn't know.

src/components/viewPartsButton.jsx
rknizzle: @ErikAmerico after you make changes based on my comments, just make a reply here saying that you did it so that I know that I can 'Resolve Conversation' just so we can keep the PR a bit cleaner and close all the conversations we're done with

I'm pretty sure that you did this one so Im going to resolve it!

rknizzle:
> @rknizzle
> 
> All is good now!!
> 
> I didn't fully understand the cleanup situation at first, that's why I copied that explanation in there in case you also didn't know.

Perfect thanks!


I pushed a couple commits that fix the model position bug and keeps good quality of the models.

This code is looking good so far! Let me know when you're ready for it to be merged in!

PR #27, opened 2024-03-14
File convention proposal

rknizzle:
I'm proposing a bit of a more concrete file and directory naming and structure convention so that there isn't really any uncertainty as we continue to build.

I'm not totally set on any of these things so let me know if theres anything here that you don't like or if you have any different ideas. You can leave any thoughts you have in comments in this PR


ErikAmerico:
I'm on board with everything described here. 

I also vote on using index.jsx over the component name as the file name avoiding duplicate names in the import path. I think it'll look cleaner too

ErikAmerico (approved):

PR #35, opened 2024-03-20
Keep header at top

ErikAmerico:
I tried a more simple approach, But was getting mixed results on different screen sizes. 

In app.jsx

`const headerRef = useRef(null);`

is passed to `<Header ref={headerRef} />`

Because the Header component is wrapped with forwardRef, it receives this prop as the second argument (idk why it automatically does that).

It then updates headerRef.current and adjusts the layout of the rest of the page to match the bottom of the header

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -10,6 +10,8 @@ function App() {
   const location = useLocation();
   const contentRef = useRef(null);
   const [isStickyFooter, setIsStickyFooter] = useState(false);
+  const headerRef = useRef(null);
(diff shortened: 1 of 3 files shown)


ErikAmerico:
the console.log in the last commit was never supposed to make it in there hahah

rknizzle:
Cool! This works well. The logic with the refs was pretty tricky to figure out what was going on. I wonder if it would be better to just place the header component in each page that uses it. I wonder if that would simplify the logic for the sizing. This works well though so I'm going to merge it in!

PR #36, opened 2024-03-21
User already exists when registering

ErikAmerico:
I believe you are sending a status code of 400 when a user already exists in the database. So i'm just checking if we got that status back and changing the error message.

I'm guessing you don't get the same eslint errors as i do (or at all?), It doesn't like the word`it's`, so I just used co-pilot's quick fix of `it&apos;s`

### DIFF
--- a/src/components/LandingPage/LandingPage.jsx
+++ b/src/components/LandingPage/LandingPage.jsx
@@ -29,7 +29,7 @@ export default function LandingPage() {
             We provide powerful software tools for analyzing 3D models and
             extracting critical data to ensure that they will be correctly
             manufactured with no issues. Gain insights and find issues before
-            it's too late.
+            it&apos;s too late.
           </div>
           <div id="description-list">
             <ul>
@@ -72,8 +72,8 @@ export default function LandingPage() {
         >
           <h2>Geometry Analysis</h2>
           <p>
-            Our Geometry Analysis is specifically designed to prepare your 3D models
-            for 3D printing or manufacturing by providing a comprehensive
+            Our Geometry Analysis is specifically designed to prepare your 3D
+            models for 3D printing or manufacturing by providing a comprehensive
             examination of their structure and integrity.
           </p>
           <ul>
--- a/src/components/loginRegister/RegisterPage/RegisterPage.jsx
+++ b/src/components/loginRegister/RegisterPage/RegisterPage.jsx
@@ -142,15 +142,28 @@ const RegisterPage = () => {
         });
       }
     } catch (error) {
-      // Registration failed
-      notification.error({
-        message: "Registration Failed",
-        description: "An error occurred during registration. Please try again.",
-        icon: <ExclamationCircleTwoTone twoToneColor="#eb2f96" />,
-        placement: "top",
-        duration: 4.5,
-        style: { width: 300 },
-      });
+      if (error.status === 400) {
+        // User already exists
+        notification.error({
+          message: "Registration Failed",
+          description: "An account with this email already exists.",
+          icon: <ExclamationCircleTwoTone twoToneColor="#eb2f96" />,
+          placement: "top",
+          duration: 4.5,
+          style: { width: 300 },
+        });
+      } else {
+        // Registration failed
+        notification.error({
+          message: "Registration Failed",
+          description:
+            "An error occurred during registration. Please try again.",
+          icon: <ExclamationCircleTwoTone twoToneColor="#eb2f96" />,
+          placement: "top",
+          duration: 4.5,
+          style: { width: 300 },
+        });
+      }
     }
   };
 
--- a/src/components/loginRegister/RegisterPage/RegisterPage.jsx
+++ b/src/components/loginRegister/RegisterPage/RegisterPage.jsx
@@ -142,7 +142,10 @@ const RegisterPage = () => {
         });
       }
     } catch (error) {
-      if (error.status === 400) {
+      if (
+        error.message ===
+        "registration failed because a user with that email already exists"
+      ) {
         // User already exists
         notification.error({
           message: "Registration Failed",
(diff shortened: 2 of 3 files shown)


src/components/loginRegister/registerRequest.js
```diff
@@ -12,7 +12,10 @@ export const register = async (email, password) => {
     });
 
     if (!response.ok) {
-      throw new Error("Registration failed");
+      throw {
+        message: "Registration failed",
+        status: response.status,
+      };
```
rknizzle: Lets only throw Error objects.
For example like this:
`throw new Error('Registration failed')`

Thats the convention for javascript developers for the most part. Using an Error object rather than just an object provides a stack trace too which can be useful sometimes for finding where an error started.
This is what a stack trace is:
js-stack-trace

rknizzle: The stuff inside this block:
```
if (!response.ok) {
  // all request errors go into this block
}
```

Is where any possible error will go to. So you cant just assume that its an error because of a user already existing with that email. Theres tons of reasons why an error could occur in this request.

So if you're looking to throw an error because of an email already existing you'll need to check for that specific error.

The response body for this specific error will look like this:
```
{
  detail: 'account with that email already exists'
}

```
and it will have a status code of 400

So your code would maybe be something like this:

```
if (!response.ok) {
  const response_body = await response.json()
  if (response_body.detail.includes('email already exists')) {
    throw new Error('registration failed because a user with that email already exists')
  }
}
```

rknizzle: Status codes are a general concept in web development and not just something that Im using. I wrote up a quick doc of how I use status codes in Propermesh:

https://github.com/rknizzle/propermesh/blob/master/docs/status_codes.md

So you can see that just checking a status_code of 400 isnt specific enough to know that its an error because of the email already existing.

src/components/loginRegister/Register.jsx
```diff
@@ -142,15 +142,28 @@ const Register = () => {
...
-        message: "Registration Failed",
-        description: "An error occurred during registration. Please try again.",
-        icon: <ExclamationCircleTwoTone twoToneColor="#eb2f96" />,
-        placement: "top",
-        duration: 4.5,
-        style: { width: 300 },
-      });
+      if (error.status === 400) {
+        // User already exists
+        notification.error({
```
rknizzle: So if you're other function throws an error something like:
```
throw new Error('registration failed because a user with that email already exists')
```

You'd maybe want to have the code in here like this:
```js
try {
  // request is sent here
} catch(err) {
  if (err.message.includes('email already exists')) {
     // show message about email already existing
  } 

  // show generic error message
}
```

rknizzle (changes requested):
Good attempt! I didn't give you enough context in the github issue to know exactly what to look for so I described the error response message better in some of my comments here.

We're getting into a bit of a tricky concept here of efficiently communicating error responses between the backend and frontend.

HTTP status codes can show the general group of the response but it doesn't give specific enough information to know exactly what the error is.

I'm using a very basic system of communicating errors by just sending a string message back from the backend for the frontend to check. This is pretty common. Some systems also make their own code system to specify specific errors.

Like my response could have like a status code of 400 and then a body of:
```
{
  "message": "user already exists with that email",
  "code": "E105"
}
```

And that code could be a custom code that is defined on the backend and it could specify this exact error. 
This is nice because you don't just have to check for the string on the frontend; you have a very specfic code to check for.
Because its possible that the string message gets changed on the backend and then the frontend would no longer correctly identify the type of error.

This is a scenario where testing can be cool. We could have a test that the frontend tries to create a user that already exists and that an error popup comes up saying that user already exists. If the backend message changes to have different wording and the frontend no longer correctly identifies the error by its string then the error message wont popup and the test will fail.

Adding custom codes like that above example is extra work though. So for now im Just sticking with string error messages in the response.

This might be kind of confusing so you can ask for clarification on any of these examples in a comment or reply in this PR

src/components/loginRegister/registerRequest.js
ErikAmerico: I am going to have to familiarize myself with the ins and outs of a stack trace. 

I see them all the time but never know how to read them

src/components/loginRegister/Register.jsx
ErikAmerico: @rknizzle 
How come you are always using `.includes` instead of requiring the error message to match something you define? 

or for `response_body.detail === "account with that email already exists"`

instead of `response_body.detail.includes('email already exists')`

In register.jsx I have 
image

which would match the error thrown in registerRequest.js

image

ErikAmerico: Is this incase you change the response_body.detail in the future? and it will most likely still contain `email already exists`?

rknizzle: Yea exactly; either way will work I just feel like this method gives a little more flexibility to maybe still work if the error message on the backend changes slightly. A change to the message on the backend would realistically be very likely to make this fail anyway though so it doesnt really matter

src/components/loginRegister/registerRequest.js
rknizzle: Yea theres usually a bunch of gibberish in them so they're hard to figure out. Usually at the top or bottom though they'll kind of show you which file the error originated from

rknizzle:
This looks to me like its working now! Good to merge in? @ErikAmerico

ErikAmerico:
@rknizzle 

Yeah for sure good to merge in!

rknizzle:
@ErikAmerico 
Ok cool! Github couldn't merge automatically so I did:
```bash
git fetch
git checkout user-already-exists-when-registering
git rebase -i origin/main
git push --force-with-lease
```

And now Ill merge it on github.

Rebase is similar to merge. It grabbed the commits in the `user-already-exists-when-registering` branch and placed them on top of main

PR #37, opened 2024-03-21
Units selector

ErikAmerico:
When a part is loaded into the viewer

The user can select `mm` or `inches` as the unit measurement, updating the db.

If a part already has units defined, those are shown by default and can be still be updated.

If a part hasn't been loaded then the user can't select a unit of measurement and there is a toolTip that explains why they can't view the select dropdown.

### DIFF
--- a/src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
+++ b/src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
@@ -4,6 +4,7 @@ import { CheckCircleOutlined } from "@ant-design/icons";
 import "./unitsSelector.css";
 import PropTypes from "prop-types";
 import { updatePartUnits } from "./updatePartUnits";
+import { getPartData } from "./getPartData";
 
 const UnitsSelector = ({ partId }) => {
   const [units, setUnits] = useState(undefined);
@@ -13,11 +14,7 @@ const UnitsSelector = ({ partId }) => {
     const getCurrentUnits = async () => {
       if (partId) {
         try {
-          const response = await fetch(`/api/v0/parts/${partId}`);
-          if (!response.ok) {
-            throw new Error("Failed to get part data.");
-          }
-          const data = await response.json();
+          const data = await getPartData(partId);
           setUnits(data.units);
         } catch (error) {
           console.error("Error getting part data:", error);
--- a/src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
+++ b/src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useState, useEffect } from "react";
 import { Select, Space } from "antd";
 import { CheckCircleOutlined } from "@ant-design/icons";
 import "./unitsSelector.css";
@@ -7,6 +7,25 @@ import PropTypes from "prop-types";
 const UnitsSelector = ({ partId }) => {
   const [units, setUnits] = useState(undefined);
   const [showSuccess, setShowSuccess] = useState(false);
+
+  useEffect(() => {
+    const getCurrentUnits = async () => {
+      if (partId) {
+        try {
+          const response = await fetch(`/api/v0/parts/${partId}`);
+          if (!response.ok) {
+            throw new Error("Failed to get part data.");
+          }
+          const data = await response.json();
+          setUnits(data.units);
+        } catch (error) {
+          console.error("Error getting part data:", error);
+        }
+      }
+    };
+
+    getCurrentUnits();
+  }, [partId]);
 
   const handleChange = async (value) => {
     setUnits(value);
@@ -41,7 +60,7 @@ const UnitsSelector = ({ partId }) => {
         <span id="units-selector-label">Units: </span>
         <Select
           placeholder="Select units"
-          defaultValue={units}
+          value={units}
           style={{ width: 120 }}
           onChange={handleChange}
           options={[
--- a/src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
+++ b/src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
@@ -3,6 +3,7 @@ import { Select, Space, Tooltip, Modal } from "antd";
 import { CheckCircleOutlined } from "@ant-design/icons";
 import "./unitsSelector.css";
 import PropTypes from "prop-types";
+import { updatePartUnits } from "./updatePartUnits";
 
 const UnitsSelector = ({ partId }) => {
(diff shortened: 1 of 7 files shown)


src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
```diff
@@ -54,20 +54,28 @@ const UnitsSelector = ({ partId }) => {
     }
   };
 
+  const isDisabled = !partId;
+  const tooltipTitle = partId
+    ? ""
+    : "View or upload a part to the viewer before selecting a unit of measurement.";
+
```
rknizzle, line 54: This tooltip is great! Is it possible to also make the unit selector less greyed out when there is a part in the page?

I feel like it looks pretty greyed out and unclickable even when their is a part.
unit-selector-red



Is that an easy change?

src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
```diff
@@ -26,8 +49,14 @@ const UnitsSelector = ({ partId }) => {
...
           ]}
         />
-      </Space>
-    </div>
+        {showSuccess && (
+          <CheckCircleOutlined
+            style={{ color: "green", fontSize: "16px" }}
+            id="unit-update-success-icon"
+          />
+        )}
```
rknizzle, line 77: I love this little graphic

src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
```diff
@@ -1,20 +1,43 @@
...
+
+      const data = await response.json();
+      console.log(data);
+
+      setShowSuccess(true);
+      setTimeout(() => {
+        setShowSuccess(false);
+      }, 2000);
+    } catch (error) {
+      console.error("Error updating units:", error);
```
rknizzle, line 38: It might be good to pop up a little message saying that it failed to update units

src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
```diff
@@ -1,20 +1,43 @@
...
+        },
+        body: JSON.stringify({ units: value }),
+      });
+
+      if (!response.ok) {
+        throw new Error("Failed to update units.");
+      }
+
+      const data = await response.json();
+      console.log(data);
```
rknizzle: I find HTTP requests to be good things to put into their own functions. Its a lot of setup code that can make the component harder to read through. I'd recommend putting this into its own function and calling it here.

Code for HTTP requests can also tend to grow a lot over time because code for handling different types of error responses might continuously be added. So its good to have it in its own function to put all of that logic and not have it fill up the React component

src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
```diff
@@ -8,6 +8,25 @@ const UnitsSelector = ({ partId }) => {
...
 
+  useEffect(() => {
+    const getCurrentUnits = async () => {
+      if (partId) {
+        try {
+          const response = await fetch(`/api/v0/parts/${partId}`);
+          if (!response.ok) {
+            throw new Error("Failed to get part data.");
+          }
+          const data = await response.json();
```
rknizzle: This can probably go into its own file too. 

I think It'll eventually move up to the AnalysisPage and then just pass the units down to this component.

src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
```diff
@@ -8,6 +8,25 @@ const UnitsSelector = ({ partId }) => {
...
+          setUnits(data.units);
+        } catch (error) {
+          console.error("Error getting part data:", error);
```
rknizzle, line 20: This would maybe be good to have an error message popup but dont worry about this one right now because I think this call might get moved to a higher component as you start building out the analysis components

rknizzle (changes requested):
Nice! This works well and I like all of the graphics a lot. 
I left a couple little suggestions in the comments

src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
rknizzle: If its not an easy change, then dont worry about this. Its not a big deal

ErikAmerico: @rknizzle yeah it should be an easy change!! 

I'll check it out this evening

src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
ErikAmerico: Makes sense!

I'll follow that practice from now on

src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
ErikAmerico: @rknizzle 
I added a little emoji  in the popup that we do not have to keep! I just wasn't sure what the next step should be for the user after being notified that the units didn't update. 

Should they be able to report an issue somehow?

src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
ErikAmerico: Just changing the color of the words "select units" to black when they are able to select a measurement cleared that up If you ask me!

rknizzle: Yup this looks awesome :+1:

src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
rknizzle: Yea good question. I just put in a little message saying to try again or report it to my email.
Here are my commits if you're curious: 
[11c9afc](https://github.com/rknizzle/propermesh-view/pull/37/commits/11c9afcbe49423ee8bfc27df83aa385234038da8)
[c1025a3](https://github.com/rknizzle/propermesh-view/pull/37/commits/c1025a3415096f0fa42b2a67d4806a9c3abd7acf)

@ErikAmerico

rknizzle: Also I think just the floating error message is good enough rather than needing to press ok to get out of the modal

rknizzle:
Ok this component is looking sick now! Merging this in

ErikAmerico:
Cool!!

I originally used the modal as the pop up, because the user had to acknowledge there was an issue by clicking the okay button.

But I like keeping everything as a  notification

PR #38, opened 2024-03-30
Geo analysis box

ErikAmerico:
@rknizzle 

First run at the Geometry Analysis box.

Let me know what you think!

I'm sure theres plenty of room for improvement.

### DIFF
--- a/src/components/AnalysisPage/GeoAnalysisBox/GeoAnalysisBox.jsx
+++ b/src/components/AnalysisPage/GeoAnalysisBox/GeoAnalysisBox.jsx
@@ -1,9 +1,16 @@
 import { Row, Col, Statistic } from "antd";
+import { useState, useEffect } from "react";
 import "./geoAnalysisBox.css";
 import StartButton from "./StartButton/StartButton";
 import PropTypes from "prop-types";
 
 const GeoAnalysisBox = ({ partId }) => {
+  const [geoData, setGeoData] = useState(null);
+
+  useEffect(() => {
+    console.log("geoData", geoData);
+  }, [geoData]);
+
   return (
     <div id="geo-analysis-box">
       <h2 id="geo-analysis-title">Geometry Analysis</h2>
@@ -23,7 +30,7 @@ const GeoAnalysisBox = ({ partId }) => {
         </Col>
         <Col span={12}>
           <Statistic title="# of Shells" value={3} precision={0} />
-          <StartButton partId={partId} />
+          <StartButton partId={partId} setGeoData={setGeoData} />
         </Col>
       </Row>
     </div>
--- a/src/components/AnalysisPage/GeoAnalysisBox/GeoAnalysisBox.jsx
+++ b/src/components/AnalysisPage/GeoAnalysisBox/GeoAnalysisBox.jsx
@@ -1,14 +1,33 @@
-import { Row, Col, Statistic } from "antd";
+import { Row, Col } from "antd";
 import { useState, useEffect } from "react";
 import "./geoAnalysisBox.css";
 import StartButton from "./StartButton/StartButton";
+import GeoStatistic from "./GeoStatistic";
 import PropTypes from "prop-types";
 
 const GeoAnalysisBox = ({ partId }) => {
   const [geoData, setGeoData] = useState(null);
+  const [volume, setVolume] = useState(null);
+  const [surfaceArea, setSurfaceArea] = useState(null);
+  const [numBoundaryEdges, setNumBoundaryEdges] = useState(null);
+  const [numShells, setNumShells] = useState(null);
+  const [roundedVolume, setRoundedVolume] = useState("--");
+  const [roundedSurfaceArea, setRoundedSurfaceArea] = useState("--");
+  const [roundedNumBoundaryEdges, setRoundedNumBoundaryEdges] = useState("--");
+  const [roundedNumShells, setRoundedNumShells] = useState("--");
 
   useEffect(() => {
-    console.log("geoData", geoData);
+    console.log(geoData);
+    if (geoData) {
+      setVolume(geoData.volume);
+      setSurfaceArea(geoData.surface_area);
+      setNumBoundaryEdges(geoData.num_boundary_edges);
+      setNumShells(geoData.num_shells);
+      setRoundedVolume(Math.round(geoData.volume));
+      setRoundedSurfaceArea(Math.round(geoData.surface_area));
+      setRoundedNumBoundaryEdges(Math.round(geoData.num_boundary_edges));
+      setRoundedNumShells(Math.round(geoData.num_shells));
+    }
   }, [geoData]);
 
   return (
@@ -16,20 +35,34 @@ const GeoAnalysisBox = ({ partId }) => {
       <h2 id="geo-analysis-title">Geometry Analysis</h2>
       <Row id="geo-top-row">
         <Col span={12}>
-          <div className="statistic-container">
-            <Statistic title="Volume" value={117.5} />
-          </div>
+          <GeoStatistic
+            title="Volume"
+            value={roundedVolume}
+            preciseValue={volume}
+          />
         </Col>
         <Col span={12}>
-          <Statistic title="# of Boundary Edges" value={12} precision={0} />
+          <GeoStatistic
+            title="# of Boundary Edges"
+            value={roundedNumBoundaryEdges}
+            preciseValue={numBoundaryEdges}
+          />
         </Col>
       </Row>
       <Row>
         <Col span={12}>
-          <Statistic title="Surface Area" value={240.7} />
+          <GeoStatistic
+            title="Surface Area"
+            value={roundedSurfaceArea}
+            preciseValue={surfaceArea}
+          />
         </Col>
         <Col span={12}>
-          <Statistic title="# of Shells" value={3} precision={0} />
+          <GeoStatistic
+            title="# of Shells"
+            value={roundedNumShells}
+            preciseValue={numShells}
+          />
           <StartButton partId={partId} setGeoData={setGeoData} />
         </Col>
       </Row>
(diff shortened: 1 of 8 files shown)


src/components/AnalysisPage/GeoAnalysisBox/StartButton/pollForResults.js
```diff
@@ -0,0 +1,17 @@
...
+  const response = await fetch(`/api/v0/geometry/${jobId}`);
+  if (!response.ok) {
+    throw new Error("Failed to fetch analysis results");
+  }
+
+  const results = await response.json();
+
+  if (results.status === "complete") {
+    setIsLoading(false);
+    setGeoData(results);
```
rknizzle: I'd recommend keeping these outside of the http request file just to keep the http request code separate from any React state setting

src/components/AnalysisPage/GeoAnalysisBox/StartButton/StartButton.jsx
```diff
@@ -22,15 +23,23 @@ const StartButton = ({ partId, setGeoData }) => {
     }
   };
 
+  const isDisabled = !partId;
+  const tooltipTitle = partId
+    ? ""
+    : "View or upload a part to the viewer before running Geometry Analysis.";
+
```
rknizzle: This is really good UI

src/components/AnalysisPage/GeoAnalysisBox/StartButton/pollForResults.js
```diff
@@ -1,17 +1,26 @@
...
 
-  if (results.status === "complete") {
-    setIsLoading(false);
-    setGeoData(results);
-    return results;
-  } else {
-    setTimeout(() => pollForResults(jobId, setIsLoading, setGeoData), 1000);
-  }
+    poll();
+  });
```
rknizzle: I'm pretty sure that this change in commit aba9dd58aa01b3d3999f1f097a0b6ce6bf1710dd isn't doing anything. Did you notice any behavior change in the app when you made this change?
pollForResults was already returning a promise automatically. You can tell by the fact that it is an 'async' function and it uses await.

Fetch returns a promise and as soon as a function awaits a promise returning function (fetch) it also returns a promise

Let me know if you need any more info on promises

src/components/AnalysisPage/GeoAnalysisBox/StartButton/StartButton.jsx
```diff
@@ -18,9 +18,10 @@ const StartButton = ({ partId, setGeoData }) => {
     try {
       const job = await startGeometryAnalysis(partId);
 
-      await pollForResults(job.id, setIsLoading, setGeoData);
-      setShowCheckmark(true);
-      setTimeout(() => setShowCheckmark(false), 5000);
+      await pollForResults(job.id, setIsLoading, setGeoData).then(() => {
+        setShowCheckmark(true);
+        setTimeout(() => setShowCheckmark(false), 2000);
+      });
```
rknizzle: await and .then do the same thing so adding the .then here didn't change anything either I dont think

src/components/AnalysisPage/GeoAnalysisBox/GeoAnalysisBox.jsx
```diff
@@ -36,28 +28,36 @@ const GeoAnalysisBox = ({ partId }) => {
...
       setNumShells(geoData.num_shells);
-      setRoundedVolume(Math.round(geoData.volume));
-      setRoundedSurfaceArea(Math.round(geoData.surface_area));
-      setRoundedNumBoundaryEdges(Math.round(geoData.num_boundary_edges));
-      setRoundedNumShells(Math.round(geoData.num_shells));
     }
   }, [geoData]);
 
+  const displayValue = (value) => {
+    if (typeof value !== "number") return value;
```
rknizzle: I found the display code a bit hard to understand because of all the type checking between numbers and strings and stuff. I have some ideas for making it simpler to understand. 

- Move the displayValue function into the GeoStatistic component rather than in the GeoAnalysisBox component.
- Start the values at null again instead of as '--'
- So now the props can just be 'number' type rather than number OR string
- In the displayValue function convert null -> '--'

So now there doesn't need to be all these checks for string or number and the GeomStatistic component is the only place that uses displayValue so it makes sense for that logic to go in there

Let me know if Im missing anything or if you were checking the typeof for a reason other than just looking for the '--'.

rknizzle (changes requested):
Ok this is looking really good so far. Theres just a couple little changes I recommended and then Ill merge this in

And then the only other big update I want to make is; I want the geometry analysis box to get automatically populated again when you load in an existing part that has already done the analysis in the past.

But for now you can just update these couple little things and then I'll merge this in and we can work on a new branch!

src/components/AnalysisPage/GeoAnalysisBox/StartButton/pollForResults.js
ErikAmerico: I don't fully understand promises. 

But prior to this commit, ` setShowCheckmark(true);` and `
        setTimeout(() => setShowCheckmark(false), 2000);` in StartButton.jsx, weren't waiting waiting for pollForResults to finish. So I was getting the loading icon and the checkmark at the same time. 

Maybe When i start fixing this, we can meet up or do a little slack chat

src/components/AnalysisPage/GeoAnalysisBox/StartButton/StartButton.jsx
ErikAmerico: same thing here. 

without the .then(), They wouldn't wait for pollForResults to finish. 

Obviously, I wasn't doing something right. So I'd like to get meet up in the next couple of days while finishing these up!

src/components/AnalysisPage/GeoAnalysisBox/StartButton/pollForResults.js
ErikAmerico: Gotchya! 

I thought we were only making sure to keep the 3d model viewer react-less. 

But all http request code should stay away from react as well, understood

src/components/AnalysisPage/GeoAnalysisBox/GeoAnalysisBox.jsx
ErikAmerico: I'll give this a try!!!

src/components/AnalysisPage/GeoAnalysisBox/StartButton/StartButton.jsx
ErikAmerico: 😎

ErikAmerico:
@rknizzle  

For the geometry analysis box to get automatically populated again when you load in an existing part that has already done the analysis in the past,

Is there something set up on the backend for me to check? Should i be checking to see if a partId is associated with a job, and then get the details of that job?

src/components/AnalysisPage/GeoAnalysisBox/StartButton/pollForResults.js
rknizzle: Yea its not a huge deal but I think its just generally a good practice to keep like 'pure' javascript stuff separate from React javascript stuff. It can just make it easier if you wanted to just test out the http request code by itself without needing to have React stuff passed in

src/components/AnalysisPage/GeoAnalysisBox/StartButton/pollForResults.js
rknizzle: Ok there could be something kind of complicated going on then. Ya lets meet up to work on this. Let me know when you do go to work on it

rknizzle:
> @rknizzle
> 
> For the geometry analysis box to get automatically populated again when you load in an existing part that has already done the analysis in the past,
> 
> Is there something set up on the backend for me to check? Should i be checking to see if a partId is associated with a job, and then get the details of that job?

@ErikAmerico 
Yea good question. You can use the `GET /api/v0/parts/{part_id}` request.
(Heres a link to it in the API docs so you can see the response format: https://propermesh.com/docs#/Parts/get_part_api_v0_parts__part_id__get) 

That request will respond with the full part along with all of its analysis data. So the geometry_analysis property in the response will either be null or it'll be that geometry analysis job data object if it has already been started. So you can make that call when the part loads, check if the geometry analysis value is not null, and then if its complete you can populate the geo box with that info. 

(theres also a chance that it can still be in_progress in which case we would want the frontend to start polling for the result again automatically, but thats not something to think about yet)

We can work on this together if you want too

src/components/AnalysisPage/GeoAnalysisBox/StartButton/pollForResults.js
```diff
@@ -1,26 +1,23 @@
...
+    const results = await response.json();
+    status = results.status;
+
+    if (status !== "complete") {
+      await delay(1000);
+    } else {
+      return results;
+    }
+  }
 };
```
rknizzle: Nice easily understandable implementation

src/components/AnalysisPage/GeoAnalysisBox/GeoStatistic.jsx
```diff
@@ -2,39 +2,60 @@ import { Statistic, Tooltip } from "antd";
...
+      return "--";
+    }
 
-  const formattedValue = typeof value === "number" ? value.toString() : value;
+    // I put Shells and Boundary Edges back into GeoStatistic to keep the "--" display consistent
+    // It's also a plus that if we need a precise value for these two in the future,
+    // We're already set up for it
+    if (title === "# of Shells" || title === "# of Boundary Edges") {
+      return value;
+    }
```
rknizzle, line 16: Oh ya true I didn't think about these fields so my idea wasn't quite as clean as I thought. Good solution though

rknizzle (approved):
Nice. All the changes look good now. Let me know if you want me to merge this PR or you can keep working in this branch

ErikAmerico:
You can merge it in and i'll start a new branch for the auto-population!

PR #39, opened 2024-04-02
Auto populate geometry box

ErikAmerico:
Auto populate the Geometry-Analysis box with data if a part has already had the analysis done in the past.

### DIFF
--- a/src/components/AnalysisPage/AnalysisPage.jsx
+++ b/src/components/AnalysisPage/AnalysisPage.jsx
@@ -10,13 +10,22 @@ const AnalysisPage = () => {
   const [fileFor3dModel, setFileFor3dModel] = useState(null);
   const [partId, setPartId] = useState(null);
   const [autoPopAnalysis, setAutoPopAnalysis] = useState(null);
+  const [analysisComplete, setAnalysisComplete] = useState(false);
+  const [showCheckmark, setShowCheckmark] = useState(false);
 
   return (
     <div>
       <h1 style={{ fontFamily: '"Roboto", sans-serif' }}>Analysis Page</h1>
       <Row>
         <Col span={6}>
-          <GeoAnalysisBox partId={partId} autoPopAnalysis={autoPopAnalysis} />
+          <GeoAnalysisBox
+            partId={partId}
+            autoPopAnalysis={autoPopAnalysis}
+            analysisComplete={analysisComplete}
+            setAnalysisComplete={setAnalysisComplete}
+            showCheckmark={showCheckmark}
+            setShowCheckmark={setShowCheckmark}
+          />
         </Col>
       </Row>
       <UploadPartButton
@@ -27,6 +36,8 @@ const AnalysisPage = () => {
         setFileFor3dModel={setFileFor3dModel}
         setPartId={setPartId}
         setAutoPopAnalysis={setAutoPopAnalysis}
+        setAnalysisComplete={setAnalysisComplete}
+        setShowCheckmark={setShowCheckmark}
       />
       <ModelViewer fileFor3dModel={fileFor3dModel} />
       <UnitsSelector partId={partId} />
--- a/src/components/AnalysisPage/AnalysisPage.jsx
+++ b/src/components/AnalysisPage/AnalysisPage.jsx
@@ -9,13 +9,14 @@ import { Row, Col } from "antd";
 const AnalysisPage = () => {
   const [fileFor3dModel, setFileFor3dModel] = useState(null);
   const [partId, setPartId] = useState(null);
+  const [autoPopAnalysis, setAutoPopAnalysis] = useState(null);
 
   return (
     <div>
       <h1 style={{ fontFamily: '"Roboto", sans-serif' }}>Analysis Page</h1>
       <Row>
         <Col span={6}>
-          <GeoAnalysisBox partId={partId} />
+          <GeoAnalysisBox partId={partId} autoPopAnalysis={autoPopAnalysis} />
         </Col>
       </Row>
       <UploadPartButton
@@ -25,6 +26,7 @@ const AnalysisPage = () => {
       <ViewPartsButton
         setFileFor3dModel={setFileFor3dModel}
         setPartId={setPartId}
+        setAutoPopAnalysis={setAutoPopAnalysis}
       />
       <ModelViewer fileFor3dModel={fileFor3dModel} />
       <UnitsSelector partId={partId} />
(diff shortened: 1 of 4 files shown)


src/components/AnalysisPage/GeoAnalysisBox/GeoAnalysisBox.jsx
```diff
@@ -27,7 +34,10 @@ const GeoAnalysisBox = ({ partId, autoPopAnalysis }) => {
       setSurfaceArea(autoPopAnalysis.surface_area);
       setNumBoundaryEdges(autoPopAnalysis.num_boundary_edges);
       setNumShells(autoPopAnalysis.num_shells);
+      setShowCheckmark(true);
+      setAnalysisComplete(true);
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
ErikAmerico, line 40: This line is just disabling a warning, react wants setShowCheckmark and setAnalysisComplete to be in the dependency array. But they shouldn't be included.

src/components/AnalysisPage/GeoAnalysisBox/StartButton/StartButton.jsx
```diff
@@ -6,17 +6,23 @@ import { pollForResults } from "./pollForResults";
...
-  const [showCheckmark, setShowCheckmark] = useState(false);
   const [showFailure, setShowFailure] = useState(false);
-  const [analysisComplete, setAnalysisComplete] = useState(false);
 
   useEffect(() => {
     setIsLoading(false);
     setShowCheckmark(false);
     setShowFailure(false);
     setAnalysisComplete(false);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
ErikAmerico, line 25: same thing here

src/components/AnalysisPage/AnalysisPage.jsx
```diff
@@ -9,13 +9,14 @@ import { Row, Col } from "antd";
 const AnalysisPage = () => {
   const [fileFor3dModel, setFileFor3dModel] = useState(null);
   const [partId, setPartId] = useState(null);
+  const [autoPopAnalysis, setAutoPopAnalysis] = useState(null);
```
ErikAmerico, line 12: When add the thickness analysis to this, we can create an object autoPopAnalysis,

`const [autoPopAnalysis, setAutoPopAnalysis] = useState({
    geometry_analysis: null,
    thickness_analysis: null,
  });`

And then I'll make adjustments so everything works alongside each other.

src/components/AnalysisPage/GeoAnalysisBox/GeoAnalysisBox.jsx
```diff
@@ -21,6 +21,15 @@ const GeoAnalysisBox = ({ partId }) => {
...
+      setVolume(autoPopAnalysis.volume);
+      setSurfaceArea(autoPopAnalysis.surface_area);
+      setNumBoundaryEdges(autoPopAnalysis.num_boundary_edges);
+      setNumShells(autoPopAnalysis.num_shells);
+    }
+  }, [autoPopAnalysis]);
+
   useEffect(() => {
     console.log(geoData);
     if (geoData) {
```
rknizzle: I wonder if it would be better to just have one geom analysis data state object in the Analysis Page that can be set by both the geom box or the list parts modal rather than having two separate states (one in the geo box and one in the analysis page) that can both hold the geom data

src/components/AnalysisPage/AnalysisPage.jsx
```diff
@@ -10,13 +10,22 @@ const AnalysisPage = () => {
   const [fileFor3dModel, setFileFor3dModel] = useState(null);
   const [partId, setPartId] = useState(null);
   const [autoPopAnalysis, setAutoPopAnalysis] = useState(null);
+  const [analysisComplete, setAnalysisComplete] = useState(false);
+  const [showCheckmark, setShowCheckmark] = useState(false);
```
rknizzle, line 14: I think that handling the showing that the geom analysis is completed can just all be handled in the geo box component rather than needing to be up in the analysis page. Whether or not a checkmark is showing on the geo box isn't really something that the AnalysisPage needs to care about in its code.

rknizzle (commented):
Cool this works perfectly!

I think that there may be a better way to organize the state in the components.

I'm thinking that maybe the geo data should be stored in the AnalysisPage so that either the geo box or the ViewPartsButton could both set and read it. Whereas right now you have both the geo data state in the GeoBox and the autoPopAnalysis in the AnalysisPage that can have the geo data stored.

And then the geo box can handle its own setting of the checkmark and completion rather than having that set by other components.

This works well though so I'm going to merge this PR in and you can just keep thinking about how the state and props are organized as you keep going. Let me know if you want to work with me to try out those state updates

src/components/AnalysisPage/GeoAnalysisBox/GeoAnalysisBox.jsx
ErikAmerico: @rknizzle 

I was planning on using `autoPopAnalysis` for both the geo-box and thickness-box. My thinking was that it would be completely separate from geoData and what will be thinknessData. 

Where geo-data is the information gained from the initial running of the analysis, so It would only need to be used and accessible to geoAnalysisBox and its start button.

When the thickness analysis box starts rolling i'll look into this more!

src/components/AnalysisPage/AnalysisPage.jsx
ErikAmerico: I must at one point have been passing these two pieces of state down to and using them in ViewPartsButton, 

But that changed along the way, and I never moved where they were defined because you're right. There's no need for them to be in the analysis page the way they're being set up and used.

src/components/AnalysisPage/GeoAnalysisBox/GeoAnalysisBox.jsx
rknizzle: Cool ya thats one way to do it. I just think with that method there are basically 2 geoData fields. And whichever one gets used just depends on if the part was just uploaded or if it was loaded from the list of parts. And so you can see like because of having 2 geoData fields you also have to have 2 basically identical useEffects to set the dimensions. So I just think it might be better to just combine them into a single geoData field which would have to live in the analysis page so that both the ViewPartsButton and the GeomBox can set it

rknizzle: But I haven't really messed with the code so that could be more complicated than it seems from just reading through. So ya you can just keep both methods in mind as you keep building stuff and see what feels better @ErikAmerico

ErikAmerico: As I'm working through the thickness analysis, I'm seeing where we differed on this. 

My original plan of having autoPopAnalysis take care of both of the analysis boxes isn't going to work as i thought it would. 

So I'm going to get rid of autoPop and then place geoData in the analysisPage like you suggested!

@rknizzle (the sensei)

Issue #1, opened 2024-01-15
Design and implement the frontend landing page 


rknizzle:
Here is the current figma sketch of the frontend on a computer
https://www.figma.com/file/6tq7J98zhIiO6n0InRwkKd/Untitled?type=design&node-id=0-1&mode=design

The design is WIP and isn't concrete right now. @ErikAmerico. If you work on this you can feel free to play around with the design if you have any ideas because I don't love the design so far but it's good enough to start working on

This can either be done in just regular HTML, CSS, and javascript or React could be used

NOTE: The entire sketch doesn't need to all fit on a computer screen... you might have to scroll to see the whole thing
NOTE: The layout should look good on computers and phones

rknizzle:
Erik finished this :+1:

PR #40, opened 2024-04-05
Thickness analysis box

ErikAmerico:
PR #1 

If there are any major issues, let a brother know.

But I should hopefully work through the finer details by the final PR of this feature.

### DIFF
--- a/src/components/AnalysisPage/AnalysisPage.jsx
+++ b/src/components/AnalysisPage/AnalysisPage.jsx
@@ -27,7 +27,7 @@ const AnalysisPage = () => {
             showCheckmark={showCheckmark}
             setShowCheckmark={setShowCheckmark}
           />
-          <ThickAnalysisBox />
+          <ThickAnalysisBox partId={partId} />
         </Col>
       </Row>
       <UploadPartButton
--- a/src/components/AnalysisPage/AnalysisPage.jsx
+++ b/src/components/AnalysisPage/AnalysisPage.jsx
@@ -14,7 +14,6 @@ const AnalysisPage = () => {
   const [geoData, setGeoData] = useState(null);
   //initialize as an array to use the spread operator in startButton.jsx
   const [thickData, setThickData] = useState([]);
-  const [partsThresholdHistory, setPartsThresholdHistory] = useState([]);
 
   return (
     <div>
@@ -37,15 +36,13 @@ const AnalysisPage = () => {
             thickData={thickData}
             //setThickData headed to start button
             setThickData={setThickData}
-            partsThresholdHistory={partsThresholdHistory}
-            setPartsThresholdHistory={setPartsThresholdHistory}
           />
         </Col>
       </Row>
       <UploadPartButton
         setFileFor3dModel={setFileFor3dModel}
         setPartId={setPartId}
-        setPartsThresholdHistory={setPartsThresholdHistory}
+        setThickData={setThickData}
       />
       <ViewPartsButton
         setFileFor3dModel={setFileFor3dModel}
--- a/src/components/AnalysisPage/AnalysisPage.jsx
+++ b/src/components/AnalysisPage/AnalysisPage.jsx
@@ -13,6 +13,7 @@ const AnalysisPage = () => {
   const [partId, setPartId] = useState(null);
   const [geoData, setGeoData] = useState(null);
   const [thickData, setThickData] = useState(null);
+  const [partsThresholdHistory, setPartsThresholdHistory] = useState([]);
 
   return (
     <div>
@@ -30,12 +31,18 @@ const AnalysisPage = () => {
               precise value available on hover
             </span>
           </div>
-          <ThickAnalysisBox partId={partId} thickData={thickData} />
+          <ThickAnalysisBox
+            partId={partId}
+            thickData={thickData}
+            partsThresholdHistory={partsThresholdHistory}
+            setPartsThresholdHistory={setPartsThresholdHistory}
+          />
         </Col>
       </Row>
       <UploadPartButton
         setFileFor3dModel={setFileFor3dModel}
         setPartId={setPartId}
+        setPartsThresholdHistory={setPartsThresholdHistory}
       />
       <ViewPartsButton
         setFileFor3dModel={setFileFor3dModel}
--- a/src/components/AnalysisPage/AnalysisPage.jsx
+++ b/src/components/AnalysisPage/AnalysisPage.jsx
@@ -44,7 +44,6 @@ const AnalysisPage = () => {
       <UploadPartButton
         setFileFor3dModel={setFileFor3dModel}
         setPartId={setPartId}
-        setListOfThicknessData={setListOfThicknessData}
       />
       <ViewPartsButton
         setFileFor3dModel={setFileFor3dModel}
--- a/src/components/AnalysisPage/AnalysisPage.jsx
+++ b/src/components/AnalysisPage/AnalysisPage.jsx
@@ -11,9 +11,7 @@ import { InfoCircleOutlined } from "@ant-design/icons";
 const AnalysisPage = () => {
   const [fileFor3dModel, setFileFor3dModel] = useState(null);
   const [partId, setPartId] = useState(null);
-  const [autoPopAnalysis, setAutoPopAnalysis] = useState(null);
-  const [analysisComplete, setAnalysisComplete] = useState(false);
-  const [showCheckmark, setShowCheckmark] = useState(false);
+  const [geoData, setGeoData] = useState(null);
 
   return (
     <div>
@@ -22,11 +20,8 @@ const AnalysisPage = () => {
         <Col span={6}>
           <GeoAnalysisBox
             partId={partId}
-            autoPopAnalysis={autoPopAnalysis}
-            analysisComplete={analysisComplete}
-            setAnalysisComplete={setAnalysisComplete}
-            showCheckmark={showCheckmark}
-            setShowCheckmark={setShowCheckmark}
+            geoData={geoData}
+            setGeoData={setGeoData}
           />
           <div style={{ marginTop: 12 }}>
             <span style={{ fontSize: "0.7rem", color: "black" }}>
@@ -44,9 +39,7 @@ const AnalysisPage = () => {
       <ViewPartsButton
         setFileFor3dModel={setFileFor3dModel}
         setPartId={setPartId}
-        setAutoPopAnalysis={setAutoPopAnalysis}
-        setAnalysisComplete={setAnalysisComplete}
-        setShowCheckmark={setShowCheckmark}
+        setGeoData={setGeoData}
       />
       <ModelViewer fileFor3dModel={fileFor3dModel} />
       <UnitsSelector partId={partId} />
--- a/src/components/AnalysisPage/AnalysisPage.jsx
+++ b/src/components/AnalysisPage/AnalysisPage.jsx
@@ -12,6 +12,7 @@ const AnalysisPage = () => {
   const [fileFor3dModel, setFileFor3dModel] = useState(null);
   const [partId, setPartId] = useState(null);
   const [geoData, setGeoData] = useState(null);
+  const [thickData, setThickData] = useState(null);
 
   return (
     <div>
@@ -29,7 +30,7 @@ const AnalysisPage = () => {
               precise value available on hover
             </span>
           </div>
-          <ThickAnalysisBox partId={partId} />
+          <ThickAnalysisBox partId={partId} thickData={thickData} />
         </Col>
       </Row>
       <UploadPartButton
@@ -40,6 +41,7 @@ const AnalysisPage = () => {
         setFileFor3dModel={setFileFor3dModel}
         setPartId={setPartId}
         setGeoData={setGeoData}
+        setThickData={setThickData}
       />
       <ModelViewer fileFor3dModel={fileFor3dModel} />
       <UnitsSelector partId={partId} />
--- a/src/components/AnalysisPage/AnalysisPage.jsx
+++ b/src/components/AnalysisPage/AnalysisPage.jsx
@@ -12,7 +12,8 @@ const AnalysisPage = () => {
   const [fileFor3dModel, setFileFor3dModel] = useState(null);
   const [partId, setPartId] = useState(null);
   const [geoData, setGeoData] = useState(null);
-  const [thickData, setThickData] = useState(null);
+  //initialize as an array to use the spread operator in startButton.jsx
+  const [thickData, setThickData] = useState([]);
   const [partsThresholdHistory, setPartsThresholdHistory] = useState([]);
 
   return (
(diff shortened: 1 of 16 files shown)


src/components/AnalysisPage/ThicknessAnalysisBox/StartButton/StartButton.jsx
```diff
@@ -1,11 +1,19 @@
...
-    console.log("Starting analysis");
-    console.log(partId);
+    console.log("Starting thickness analysis...");
+    try {
+      const job = await startThicknessAnalysis(partId, thresholdValue);
+
+      console.log(job);
+    } catch (error) {
+      console.error("Failed to start thickness analysis.", error);
+      return;
```
rknizzle: This log is basically just going print the same message twice because your startThicknessAnalysis function throws the message 'Failed to start thickness analysis' and then you log that same message here.

src/components/AnalysisPage/ThicknessAnalysisBox/StartButton/StartButton.jsx
```diff
@@ -9,7 +10,9 @@ const StartButton = ({ partId, thresholdValue }) => {
     try {
       const job = await startThicknessAnalysis(partId, thresholdValue);
 
-      console.log(job);
+      const results = await pollForResults(job.id);
+
+      console.log(results);
     } catch (error) {
       console.error("Failed to start thickness analysis.", error);
```
rknizzle: This message is also no longer necessarily accurate as of commit ddefcce106f731a3144dbd4c2391a8fab3e7f89d because the message says 'Failed to start thickness analysis' but it could also be pollForResults that throws the error

src/components/AnalysisPage/DisplayStatistic.jsx
```diff
@@ -2,15 +2,12 @@ import { Statistic, Tooltip } from "antd";
...
     if (value === null) {
       return "--";
     }
 
-    // I put Shells and Boundary Edges back into GeoStatistic to keep the "--" display consistent
-    // It's also a plus that if we need a precise value for these two in the future,
-    // We're already set up for it
     if (title === "# of Shells" || title === "# of Boundary Edges") {
       return value;
     }
```
rknizzle: I don't love that we have to specify the specific fields that are non-decimal because we'll have to keep this if statement always up to date as we add new fields to display. I'm not quite sure how to make it better though. Maybe splitting out to a separate displayStatistic for non-decimal values. Im not sure just thinking out loud

src/components/AnalysisPage/GeoAnalysisBox/GeoAnalysisBox.jsx
```diff
@@ -3,7 +3,7 @@ import { useState, useEffect } from "react";
 import "./geoAnalysisBox.css";
 import StartButton from "./StartButton/StartButton";
 import DisplayStatistic from "../DisplayStatistic";
-import { InfoCircleOutlined } from "@ant-design/icons";
+// import { InfoCircleOutlined } from "@ant-design/icons";
```
rknizzle: You might delete this in a later commit but if not; remove this commented out line

src/components/AnalysisPage/AnalysisPage.jsx
```diff
@@ -12,6 +12,7 @@ const AnalysisPage = () => {
   const [fileFor3dModel, setFileFor3dModel] = useState(null);
   const [partId, setPartId] = useState(null);
   const [geoData, setGeoData] = useState(null);
+  const [thickData, setThickData] = useState(null);
```
rknizzle: These names could end up a little confusing since they are basically the same but geoData is just an object whereas thickness is an array of objects. IDK if just adding like listOfThicknessData or something would make it clearer

src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
```diff
@@ -63,6 +67,19 @@ const ThickAnalysisBox = ({ partId, thickData }) => {
...
+          <Row>
+            <Segmented
+              options={partsThresholdHistory.map((threshold) => ({
+                label: `${threshold}`,
+                value: threshold,
+                key: `${threshold}`,
+              }))}
+            />
+          </Row>
+        </div>
```
rknizzle: highlight-one-thickness

I like how the box gets highlighted when it gets moused over so it looks clickable. I eventually want to do this same visual to the view parts rows I think so that the row looks clickable

src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
```diff
@@ -1,21 +1,29 @@
...
   const [isThin, setIsThin] = useState(null);
 
   console.log(thickData);
 
-  const partsThresholdHistory = thickData
-    ? thickData.map((data) => data.threshold)
-    : [];
+  useEffect(() => {
+    if (thickData) {
+      setPartsThresholdHistory(thickData.map((data) => data.threshold));
```
rknizzle: Can the thresholdHistory stay as a piece of state in ThicknessAnalysisBox rather than having to be stored up in AnalysisPage? I think that would be cleaner if so.

rknizzle (commented):
Very sick! The structure of state is looking pretty good now.

I don't have any explicit changes for you to make. Its looking good. I just added a few suggestions and questions to the comments

src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
rknizzle: And actually do we even need threhsoldHIstory? Or can we get the same info by just using thickness data state?

ErikAmerico: Yeah we can do that!!!

I think after my latest commit on the thickness continuation branch, it opened up all the possibilities .

Because we immediately have the latest threshold options available.

I just quickly tried it and it worked

ErikAmerico:
Ha.

I still have to create a second row for the threshold values to hangout in.

But I think this is looking pretty good otherwise.

Unless I'm missing something!

And obviously still have to make everything responsive and reorganize things across the entire analysis page

ErikAmerico:
Actually, Is the "(mm)" next to the threshold input supposed to be updated with either inches or mm depending on what is selected from the units selector?

rknizzle:
> Actually, Is the "(mm)" next to the threshold input supposed to be updated with either inches or mm depending on what is selected from the units selector?

No its always specified in mm so you're good

src/components/AnalysisPage/UploadPartButton/UploadPartButton.jsx
```diff
@@ -33,7 +29,7 @@ const UploadPartButton = ({
       if (status === "done") {
         message.success(`${info.file.name} file uploaded successfully.`);
         setPartId(info.file.response.id);
-        setPartsThresholdHistory([]);
+        setThickData([]);
```
rknizzle: TODO: Is setting the thickness data back to an empty array necessary in the UploadPartButton? Why doesnt the geom analysis data have to be set back to null here?

src/components/AnalysisPage/DisplayStatistic.jsx
```diff
@@ -8,7 +8,8 @@ const DisplayStatistic = ({ title, value }) => {
       return "--";
     }
 
-    if (title === "# of Shells" || title === "# of Boundary Edges") {
+    const valueString = value.toString();
+    if (!valueString.includes(".")) {
```
rknizzle, line 12: Good change! I didn't llike how we were checking for specific titles before. This is much easier to maintain if we add new fields

src/components/AnalysisPage/ThicknessAnalysisBox/StartButton/StartButton.jsx
```diff
@@ -55,10 +56,21 @@ const StartButton = ({
...
+    tooltipTitle =
+      "View or upload a part to the viewer before running Thickness Analysis.";
+  } else if (thresholdValue === null && units === null) {
+    tooltipTitle =
+      "Enter a threshold value and select units of measurement to start analysis";
+  } else if (thresholdValue === null) {
+    tooltipTitle = "Enter a threshold value to start analysis";
+  } else if (units === null) {
+    tooltipTitle = "Select units of measurement to start analysis";
+  }
```
rknizzle, line 73: This is a nice message in the UI :+1:

src/components/AnalysisPage/ThicknessAnalysisBox/DecimalInput.jsx
```diff
@@ -0,0 +1,31 @@
...
+  const handleChange = (e) => {
+    const newValue = e.target.value;
+    const reg = /^(\d+)?(\.\d{0,2})?$/;
+
+    if (newValue === "" || reg.test(newValue)) {
+      setInputValue(newValue);
+      const numericValue = newValue === "" ? null : parseFloat(newValue);
+      onChange(numericValue);
+    }
+  };
```
rknizzle: Add a comment to this saying that it only allows numbers and limits the the input to 2 decimal spots. I would have no idea what this is doing if it weren't for the commit message lol

rknizzle: Also I found a small bug. If you just type `.` as the first value in the input. It will display it as NaN (not a number)
nan-bug

rknizzle (changes requested):
Cool this interface feels good! I think this is just about good to merge in and we can move onto the next feature which is viewing the thickness analyzed colored model in the viewer after the thickness analysis.

I just posted a few small changes and suggestions. Also leave a reply to my previous reviews comments if you've seen those and made changes on those so I know that you've gone through all the reviews

Heres some basic info for the next feature:
https://github.com/rknizzle/propermesh-view/issues/41

@ErikAmerico oh and heres a little example me and chatGPT made of a ply viewer to test out this concept:
https://github.com/rknizzle/propermesh/blob/master/etc/ply-viewer/main.js

src/components/AnalysisPage/ThicknessAnalysisBox/DecimalInput.jsx
ErikAmerico: I briefly saw that bug last night buut couldn't replicate it!!

Nice find, nice find

src/components/AnalysisPage/ThicknessAnalysisBox/StartButton/StartButton.jsx
ErikAmerico: all good here now!

src/components/AnalysisPage/ThicknessAnalysisBox/StartButton/StartButton.jsx
ErikAmerico: speaking the truth. i took care of this

src/components/AnalysisPage/DisplayStatistic.jsx
ErikAmerico: you already saw the solution to all of these that you made comments on. but this is also taken care of hahah

src/components/AnalysisPage/AnalysisPage.jsx
ErikAmerico: I changed this to listOfThicknessData!

My bad for not commenting on these immediately after fixing them.

src/components/AnalysisPage/UploadPartButton/UploadPartButton.jsx
ErikAmerico: No it was not necessary, 

It was definitely an oversight by me, just vibing last night. 

I should have set it back to an empty array in the useEffect hook with partId as a dependency in ThickAnalysisBox.jsx. It has the same effect but less passing of states.

I fixed it!

rknizzle: Ok cool good call :+1: 
Where did you fix it?

rknizzle:
@ErikAmerico I forgot to mention I'm having the issue again where the app isnt running on my computer without me modifying imports.

So like you're importing the thickness box css file starting with an upper case letter:
issue

But the CSS file in git doesn't actually start with an upper case letter.

Same issue with the StartButton in the thickness box

ErikAmerico:
> @ErikAmerico I forgot to mention I'm having the issue again where the app isnt running on my computer without me modifying imports.
> 
> So like you're importing the thickness box css file starting with an upper case letter: issue
> 
> But the CSS file in git doesn't actually start with an upper case letter.
> 
> Same issue with the StartButton in the thickness box

Thats strange, I don't have that issue, But i do see on my computer that the import of the css files are capitalized. 

I checked back in my commits and the imports have always been that way too. 

I'll push that fix up tonight.

src/components/AnalysisPage/UploadPartButton/UploadPartButton.jsx
ErikAmerico: I removed setListOfThicknessData([]); (originally setThickData([])) from 
image
 in UploadPartButton.jsx. also removed the passing of the prop.

and just called it in the useEffect hook in ThickanalysisBox.jsx that has partId has a dependency:
image

this is exactly how geoBox is resetting geoData. 

I think I was in some sort of zone last night. Making myself do extra work hahah

rknizzle: Oh ok sweet I just couldn't find that change in the code. Did you push that commit yet?

ErikAmerico:
@rknizzle 

The css import fixes have arrived

src/components/AnalysisPage/UploadPartButton/UploadPartButton.jsx
rknizzle: Ok there it is I see the commit now! This is all good

rknizzle:
@ErikAmerico Incredible so now its just the NaN bug and I'll merge this in

src/components/AnalysisPage/ThicknessAnalysisBox/DecimalInput.jsx
ErikAmerico: @rknizzle I pushed up a fix for the NaN bug!

seems to be working.

The only thing I don't like is that if you have a value like "0.09" and go to delete just the 9. It changes the value to just "0" instead of "0.0".

Not that big of a deal I don't think. But still.

Also, you can definitely merge this in. 
But if there are too many values in the threshold history row, they will eventually start to leave the container.

image


I can take care of this when I do the responsiveness and layout of the analysis page as a whole. So we can move on if you'd like

ErikAmerico: I actually hate that you can't only delete the second decimal spot.

I'll eventually fix that too,

rknizzle: Cool ya lets try to get that fixed at some point but I'm going to merge this in and we can move onto thickness visualization viewing and the layout of the components in the page

I created an 'issue' so we can go back to this later:
https://github.com/rknizzle/propermesh-view/issues/42

PR #45, opened 2024-04-15
Analysis page styling

rknizzle:
TODO


src/App.jsx
```diff
@@ -11,37 +11,42 @@ function App() {
...
   useEffect(() => {
     const contentHeight = contentRef.current.getBoundingClientRect().height;
-    const viewportHeight = window.innerHeight;
-    // Determine if the footer should be sticky
-    // sets isStickerFooter to true if the content height is less than the viewport height, false otherwise
-    setIsStickyFooter(contentHeight < viewportHeight);
-  }, [location.pathname]); // Rerun this effect if the route changes
+    // 84 is the height of the footer across all screen sizes, using forwardRef caused
+    // the footer to always be sticky for some reason. This is a hacky fix for now
+    const workingHeight = contentHeight + headerHeight + 84;
```
rknizzle: If the 84 gets changed then this will probably break. One strategy for this is to pull the variable out into a central location like a new file where it has

`const FOOTER_HEIGHT = 84` 
(the convention for naming constant values that never change throughout the life of the program is to have them all upper case with underscores)

And then you can import that into all the places that use it so that you can just change that one spot and it'll update all the places that use it. BUT Im guessing that the other file that uses this 84 value is a css file and so I dont know if you can import variables into that file so this strategy might not work. If thats not possible then just leave a comment above both places that use the 84 value saying if you change that value then you need to go to the other place and change it there as well.

src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
```diff
@@ -96,9 +110,13 @@ const ThickAnalysisBox = ({
...
       {listOfThicknessData && (
-        <div id="listOfThicknessData-container">
-          <Row wrap={true}>
+        <div
+          ref={listOfThicknessDataContainerRef}
+          id="listOfThicknessData-container"
+        >
+          <Row wrap={true} gutter={[0, 5]}>
             {listOfThicknessData.map((data) => (
+              //not sure about having the key be the threshold value
```
rknizzle: What does the key even do?

src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
```diff
@@ -61,6 +62,19 @@ const ThickAnalysisBox = ({
...
+    const timer = setTimeout(() => {
+      const container = listOfThicknessDataContainerRef.current;
+      if (container && container.scrollHeight > container.clientHeight) {
+        container.scrollBy({ top: 100, behavior: "smooth" });
+        setTimeout(() => {
+          container.scrollBy({ top: -100, behavior: "smooth" });
+        }, 800);
+      }
+    }, 1000);
+    return () => clearTimeout(timer);
```
rknizzle, line 89: What are these 2 setTimeouts doing? Is it just adding an animation type effect?

src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
```diff
@@ -115,22 +115,24 @@ const ThickAnalysisBox = ({
...
-                      key: `${data.threshold}`,
-                    },
-                  ]}
-                  value={selectedThreshold}
-                  onChange={getSpecificDataRegardingThreshold}
-                />
-              </Col>
-            ))}
+            {listOfThicknessData
+              .filter((data) => data.units === units)
```
rknizzle: slick

src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
```diff
@@ -89,6 +89,16 @@ const ThickAnalysisBox = ({
...
 
+  const isDisabled = units === null;
+
+  let tooltipTitle = "";
+
+  console.log(isDisabled);
+  if (isDisabled) {
+    tooltipTitle =
+      "Select units of measurement before entering a threshold value";
+  }
```
rknizzle: Should this be in a useEffect or something rather than just floating in the component? I'm guessing that this will run for every render or something rather than just when the units change but I don't really know how react works. 

Also is there a way that this logic could be put into the threhsold box component rather than in the ThickAnalysisBox? I dont know which one would be better so I'm not currently implying that you should do that but im just curious if you considered that too

src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
```diff
@@ -7,26 +7,14 @@ import {
...
-          const data = await getPartData(partId);
-          setUnits(data.units);
-        } catch (error) {
-          console.error("Error getting part data:", error);
-        }
-      }
-    };
-
-    getCurrentUnits();
+    setUnits(null);
```
rknizzle: I'm looking at commit dae2d2c28dc02738489bb99c62cb0fd2483af3dd

I actually think that the units should be populated with whatever units the part is set to in the database currently. Because for each part there is going to be a 'correct' units for that part because its whatever units the designer intended the part to be modeled in. So when a part is loaded in it should bet set to those units. The only real use case for someone doing an analysis with both types of units is just if they accidentally selected the wrong one first or are curious about the size in the other units or something but there is generally just going to be one correct units for the part and so after they set that type of units then the part should be set to that units again when it gets loaded back in.
So we shouldn't be setting the units always back to null when a part is loaded in BUT now that I'm reading this code again I also don't think that the unit selector should be making the HTTP request to get the part data and then set the units there. I'm pretty sure that the View parts modal is already making that call when a part is selected and so it should probably just save the units after that call and then the unit selector can get the units passed down to it and put that value in the unit selector when the part is loaded in.

Does that make sense?

rknizzle (commented):
@ErikAmerico I created a PR for this branch so that I can review it

I like the behavior of the analysis page. Just besides the setting of the units to null when a new part gets loaded in. I have more info on that in a comment. But otherwise the behavior of the page is feeling really good to me. So all the other comments are just code suggestions and questions. But ya this is looking good to me!

Oh ya and so whats the problem with the firefox and the thickness box scrolling? Does firefox just not show the scroller on the side? I think it would probably be good to get that fixed since I use firefox. Im guessing most people will not use firefox but i think it would still be good to solve if its not too complicated

src/App.jsx
ErikAmerico: This is the only file where 84 is used. The footer just happened to end up at 84px high after we styled and added the logo. It remains that height on all screen sizes. But in the footer.css file, if we change the padding of .footer-container or height of #footer-logo, then the footer itself will change from 84px. i'll leave a comment in the css saying if we were to ever change those values, then we have to check what height that changed the footer to and then adjust it in app.jsx

ErikAmerico: I'll leave a comment in app.jsx as well

src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
ErikAmerico: Without the key, I get an eslint warning saying that I'm missing a key prop for the element in iterator. 

data.threshold is the key given to its own item, so it will always be unique. Now that i'm looking into it more, data.threshold actually makes the most sense in this situation I think

The keys help react determine which items have been added or deleted.

src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
ErikAmerico: Yeah they are for the scrolling animation of the threshold history container when there is more than 1 row. 

it waits a second to start the animation, then the scroll bar moves to the bottom while the setTimeout is waiting 800ms to run and bring the scroll bar to its original spot

rknizzle: Ok sweet thats probably cool. Is it possible to get it to happen right now? It looks like this gets triggered when a new part is loaded in but the analysis page currently loads in the part with no units selected so I dont think this will ever happen. Although I know that I made a comment below to change it so that parts are loaded in with units so this might work again after you change that. But also should this useEffect be triggered when the units are changed now not when the partId is changed since the thresholdHistory depends on the units selected for a part?

src/App.jsx
rknizzle: Cool! Can we also just lock the footer to that height explicitly rather than the height being determined by the logo size or padding?  I'm not really sure how this works

src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
ErikAmerico: Yeah that makes sense! i can put it in a useEffect with partId and isDisabled as dependencies. Then create a 
`const [tooltipTitle, setTooltipTitle] = useState("");` state and set the state in the useEffect.

To put it in the threshold box component (decimalInput.jsx), I'll just have to pass down units and partId to it as a prop so we it will always know if there are units selected and when a part changes.

src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
ErikAmerico: > Because for each part there is going to be a 'correct' units for that part because its whatever units the designer intended the part to be modeled in. So when a part is loaded in it should bet set to those units. The only real use case for someone doing an analysis with both types of units is just if they accidentally selected the wrong one first or are curious about the size in the other units or something but there is generally just going to be one correct units for the part and so after they set that type of units then the part should be set to that units again when it gets loaded back in.
> 

Ooooohh, see I didn't understand that this was the workflow. I kept thinking that it was annoying that it was just automatically set to a unit, not giving the person a fresh slate to choose from the beginning.

src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
ErikAmerico: I just tested it out and you're correct, It'll never happen. 

But after I make the change for the parts to be loaded with units, I think maybe we would want this effect to run when units change and also when a new part is loaded.. 

Because i'm thinking that maybe if a new part has the same units as the previous part it might not run because nothing actually changed? 

or maybe it will run because units gets set to null for a brief second. 

The answer is coming soon

ErikAmerico: Yeah this works and it works better.

Before if you switched between mm and inches and they both had more than 1 row. you wouldn't get the effect on that switch. Now you get the effect on the load and the switch even if the previous part had the same units

ErikAmerico:
@rknizzle oh i think
```
 useEffect(() => {
    const timer = setTimeout(() => {
      const container = listOfThicknessDataContainerRef.current;
      if (container && container.scrollHeight > container.clientHeight) {
        container.scrollBy({ top: 100, behavior: "smooth" });
        setTimeout(() => {
          container.scrollBy({ top: -100, behavior: "smooth" });
        }, 800);
      }
    }, 1000);
    return () => clearTimeout(timer);
  }, [units]);
```

was only working with just units as a dependency because I was still setting units to null in unitsSelector. after i removed that useEffect hook in my latest commit. The scroll effect doesn't recognize a change when a part changes but has the same units as the part before it. 

adding partId in the dependency array brings us back to good

src/components/AnalysisPage/UnitsSelector/UnitsSelector.jsx
ErikAmerico: viewpartsbutton is setting the units now!

ErikAmerico:
> Oh ya and so whats the problem with the firefox and the thickness box scrolling? Does firefox just not show the scroller on the side? I think it would probably be good to get that fixed since I use firefox. Im guessing most people will not use firefox but i think it would still be good to solve if its not too complicated

It shows the scroll bar, just not in the exact style that you'd see on chrome or safari. 

when you hover over the scroll 'thumb' ( i think thats what its called, the little lever), on chrome and safari, it turns purple.
when you hover on firefox, it gets slightly wider, and the background color changes to white. doesn't look great

Also it disappears when you're not scrolling on firefox. with chrome and safari its always visible..

I need to google how to have multiple css files or something for different kinds of browsers

src/App.jsx
ErikAmerico: yeah thats probably better than adjusting the size of the footer depending on what we want to do with the content of the footer. 

we can set the footer height to 84, and make the content work around that

rknizzle:
> > Oh ya and so whats the problem with the firefox and the thickness box scrolling? Does firefox just not show the scroller on the side? I think it would probably be good to get that fixed since I use firefox. Im guessing most people will not use firefox but i think it would still be good to solve if its not too complicated
> 
> It shows the scroll bar, just not in the exact style that you'd see on chrome or safari.
> 
> when you hover over the scroll 'thumb' ( i think thats what its called, the little lever), on chrome and safari, it turns purple. when you hover on firefox, it gets slightly wider, and the background color changes to white. doesn't look great
> 
> Also it disappears when you're not scrolling on firefox. with chrome and safari its always visible..
> 
> I need to google how to have multiple css files or something for different kinds of browsers

I feel like my firefox doesn't behave like you describe. Mine looks good but it just doesn't have a scroll bar at all so its kind of hard to tell that you can scroll it there. The animation helps see that its scrollable though.
thresho-hist

rknizzle:
@ErikAmerico This is looking really good. I'm good to merge this whenever you are. Just let me know

Also can you give my PR a review when you get a chance? (feel free to just say it looks good if you don't see any problems or you can ask questions or whatever in there)
https://github.com/rknizzle/propermesh-view/pull/46

ErikAmerico:
> > > Oh ya and so whats the problem with the firefox and the thickness box scrolling? Does firefox just not show the scroller on the side? I think it would probably be good to get that fixed since I use firefox. Im guessing most people will not use firefox but i think it would still be good to solve if its not too complicated
> > 
> > 
> > It shows the scroll bar, just not in the exact style that you'd see on chrome or safari.
> > when you hover over the scroll 'thumb' ( i think thats what its called, the little lever), on chrome and safari, it turns purple. when you hover on firefox, it gets slightly wider, and the background color changes to white. doesn't look great
> > Also it disappears when you're not scrolling on firefox. with chrome and safari its always visible..
> > I need to google how to have multiple css files or something for different kinds of browsers
> 
> I feel like my firefox doesn't behave like you describe. Mine looks good but it just doesn't have a scroll bar at all so its kind of hard to tell that you can scroll it there. The animation helps see that its scrollable though. thresho-hist

weird, I downloaded firefox to test it out hahah but if you're cool with it then i'm cool with it!

ErikAmerico:
@rknizzle yeah you can merge it in!

I'll make a new branch for responsiveness. 

I'll review your other PR right now

ErikAmerico (approved):
I'm good with how the page looks!

PR #46, opened 2024-04-16
Fix threshold input bug

rknizzle:
Made a fix for the threshold input bug so it feels a bit smoother. I don't love this solution because now the thresholdValue is sometimes used as a string and sometimes as a number so its a bit weird but this was the best I could come up with to get the UI to feel smooth

### DIFF
--- a/src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
+++ b/src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
@@ -45,13 +45,13 @@ const ThickAnalysisBox = ({
   const getSpecificDataRegardingThreshold = (value) => {
     //get the data from listOfThicknessData that matches the threshold value and units value
     const data = listOfThicknessData.find(
-      (data) => data.threshold === value && data.units === units //**added units check
+      (data) => Number(data.threshold) == Number(value) && data.units === units //**added units check
     );
     if (data) {
       setThinSurfaceArea(data.thin_surface_area);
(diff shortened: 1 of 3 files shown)


src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
```diff
@@ -45,21 +45,21 @@ const ThickAnalysisBox = ({
   const getSpecificDataRegardingThreshold = (value) => {
     //get the data from listOfThicknessData that matches the threshold value and units value
     const data = listOfThicknessData.find(
-      (data) => data.threshold === value && data.units === units //**added units check
+      (data) => Number(data.threshold) == Number(value) && data.units === units //**added units check
     );
```
ErikAmerico, line 49: is `Number(data.threshold)` converting what potentially is a string, back to a number?

ErikAmerico (approved):
This looks good to me.

rknizzle strikes again

src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
rknizzle: Yea I think that data.threshold might always be a number whereas `value` is potentially a string but converting them both to numbers if just to be safe I guess.

And I want them both to be numbers for this comparison because if they are potentially a string then something like ".7" will not equal "0.7" but when they are both converted to numbers then they will be equal

PR #49, opened 2024-04-23
fix units selection bug

ErikAmerico:
When you uploaded a new part, the units selector would still have the value of the previous part.
Same issue for when viewing a part that hadn't had a unit selected yet, It would show the units for the previous part. 
So it looked like you could run a thickness analysis but it would fail because you didn't actually select anything. 

My watch is showing some A.M hours, so this might not have been the best way to go about this, let me know if you have a better solution!


rknizzle:
Good catch nice! Merging this in

PR #50, opened 2024-04-24
Analysis page responsiveness

ErikAmerico:
This is a decent first go at the responsiveness. 

I haven't tweaked anything for the extra large screens yet, But for laptops down to small cell phones it should be good!

There are still issues with threeJS that have to be figured out, The canvas itself doesn't resize properly with the model-viewer-container. 

There are a couple merge conflicts also, but I'll take care of those tomorrow

I tried to fix the conflicts, But I messed something up. Not all of the code from the "units selection bug fix" was merged in.

I'll probably have to undo that last "[Merge branch 'main' into analysis-page-responsiveness]" commit

### DIFF
--- a/src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
+++ b/src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
@@ -8,6 +8,7 @@ import "./3dModelViewer.css";
 const ModelViewer = ({ fileFor3dModel }) => {
   const canvasRef = useRef(null);
   const [objectURL, setObjectURL] = useState(null);
+  const [, forceContainerResize] = useState(false);
 
   useEffect(() => {
     if (fileFor3dModel) {
@@ -29,6 +30,15 @@ const ModelViewer = ({ fileFor3dModel }) => {
     // Call cleanup function when the component unmounts/before re-running the effect
     return () => cleanup();
   }, [objectURL]);
+
+  useEffect(() => {
+    const handleResize = () => {
+      forceContainerResize((prev) => !prev);
+    };
+
+    window.addEventListener("resize", handleResize);
+    return () => window.removeEventListener("resize", handleResize);
+  }, []);
 
   return (
     <div id="model-viewer-container">
--- a/src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
+++ b/src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
@@ -30,7 +30,11 @@ const ModelViewer = ({ fileFor3dModel }) => {
     return () => cleanup();
   }, [objectURL]);
 
-  return <canvas ref={canvasRef} id="model-canvas" />;
+  return (
+    <div id="model-viewer-container">
+      <canvas ref={canvasRef} id="model-canvas" />
+    </div>
+  );
 };
 
 ModelViewer.propTypes = {
--- a/src/components/AnalysisPage/AnalysisPage.jsx
+++ b/src/components/AnalysisPage/AnalysisPage.jsx
@@ -20,7 +20,7 @@ const AnalysisPage = () => {
   return (
     <div>
       <Row gutter={16}>
-        <Col span={6}>
+        <Col xs={{ span: 24, order: 4 }} lg={{ span: 6, order: 1 }}>
           <div id="geo-hoverInfo-thick-container">
             <GeoAnalysisBox
               partId={partId}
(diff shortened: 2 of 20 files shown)


src/components/AnalysisPage/AnalysisPage.jsx
```diff
@@ -63,15 +63,13 @@ const AnalysisPage = () => {
...
         >
-          <div id="view-parts-button-container">
-            <ViewPartsButton
-              setFileFor3dModel={setFileFor3dModel}
-              setPartId={setPartId}
-              setGeoData={setGeoData}
-              setListOfThicknessData={setListOfThicknessData}
-              setUnits={setUnits}
-            />
-          </div>
```
rknizzle: Im looking at commit 3b05fce9f698166d3e8753dab360bb6b68f0a8a6 and I like this pattern of having the div containers in the lower level components rather than all in the higher level AnalysisPage. This should make the css files a lot cleaner by having everything close to the component that they are styling rather than having all the CSS up in the AnalysisPage. It makes the AnalysisPage componenet cleaner too. Good idea

src/components/AnalysisPage/ThicknessAnalysisBox/StartButton/StartButton.jsx
```diff
@@ -71,11 +74,19 @@ const StartButton = ({
...
 
   function isValidThresholdInput(threshold) {
-    if (threshold === "." || threshold === ".0" || threshold === "" || threshold === "0" || threshold === "0.0" || threshold === '0.' || threshold === null) {
-      return false
+    if (
+      threshold === "." ||
+      threshold === ".0" ||
+      threshold === "" ||
+      threshold === "0" ||
+      threshold === "0.0" ||
```
rknizzle, line 82: Totally unrelated to this PR but I just realized I could probably also add '0.00' and '.00' to this list as well so that the button isnt clickable for that value

src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
```diff
@@ -30,6 +31,15 @@ const ModelViewer = ({ fileFor3dModel }) => {
     return () => cleanup();
   }, [objectURL]);
 
+  useEffect(() => {
+    const handleResize = () => {
+      forceContainerResize((prev) => !prev);
```
rknizzle, line 36: I dont really understand this. What does forceContainerResize do? Is it just placeholder code for the future? Doesn't really seem like its doing anything currently

src/components/Header/Header.jsx
```diff
@@ -3,6 +3,7 @@ import { useNavigate } from "react-router-dom";
 import "./header.css";
 import { useAuth } from "../../utils/useAuth";
 import { useEffect, forwardRef } from "react";
+import { MenuOutlined } from "@ant-design/icons";
```
rknizzle, line 6: Does the hamburger dropdown work? I can't really figure out how to get it to work in my firefox browser. Maybe just send me a video when you get a chance of it working on your system

rknizzle (commented):
Unreal
Looks really good.

I don't have much experience with styling so I don't really have anything to comment for that stuff but its definitely looking really good. No changes needed, I just posted a couple questions in my review.

We can work through the merge conflicts together if you want

And then yea we may need to update the camera or something on the threejs viewer when the size changes

src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
ErikAmerico: It was making the model viewer responsive at some point in my endeavour. But apparently at some point along the way, I adjusted things enough so that it is no longer doing anything.

I wanted to add it in its own commit, so i removed it, and then added it after the fact without double checking that it was still actually needed. I may have even wrote it wrong the second time.

I'll remove it!

ErikAmerico: I need to remove that merge commit before pushing this change. I'm about to go back into work, we'll have to do this later!

src/components/Header/Header.jsx
ErikAmerico: https://github.com/rknizzle/propermesh-view/assets/111093132/d07e9b07-ecdc-4fdf-ad52-b8aa7ff71b51

ErikAmerico: There might be something weird about it, i'll look into later. It was working fine last night

src/components/AnalysisPage/ThicknessAnalysisBox/StartButton/StartButton.jsx
rknizzle: Made a PR here for this: https://github.com/rknizzle/propermesh-view/compare/extra-zero-cases?expand=1

src/components/Header/Header.jsx
rknizzle: Ok sick! I think I just don't really know how to use the phone emulator in my browser lol

rknizzle: I think I'd prefer the options to be all or nothing. So like either all the options on the header or all of them in the burger menu rather than being able to have some on the header and some in the menu.

Super low priority though we can definitely merge this in before doing that. I just documented it in this issue: https://github.com/rknizzle/propermesh-view/issues/52

src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
rknizzle: Gotcha makes sense!

src/components/Header/Header.jsx
ErikAmerico: roger that!

PR #56, opened 2024-04-26
Redirect to login

ErikAmerico:
The ProtectedRoute component will wait for confirmLogin to finish and then determine if you're allowed to visit the analysis page.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3,7 +3,10 @@
   <head>
     <meta charset="UTF-8" />
     <link rel="icon" type="image/svg+xml" href="assets/favicon.png" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1.0, user-scalable=no"
+    />
     <title>propermesh</title>
   </head>
   <body>
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -3,7 +3,6 @@ import { Outlet } from "react-router-dom";
 import Header from "./components/Header/Header";
 import Footer from "./components/Footer/Footer";
 import { useLocation } from "react-router-dom";
-import { AuthProvider } from "./utils/authContext";
 import { useEffect, useRef, useState } from "react";
 
 function App() {
@@ -41,15 +40,13 @@ function App() {
   }, []);
 
   return (
-    <AuthProvider>
-      <>
-        <Header ref={headerRef} />
-        <div ref={contentRef} style={{ marginTop: `${headerHeight}px` }}>
-          <Outlet />
-        </div>
-        <Footer sticky={isStickyFooter} />
-      </>
-    </AuthProvider>
+    <>
+      <Header ref={headerRef} />
+      <div ref={contentRef} style={{ marginTop: `${headerHeight}px` }}>
+        <Outlet />
+      </div>
+      <Footer sticky={isStickyFooter} />
+    </>
   );
 }
 
--- a/src/components/Header/header.css
+++ b/src/components/Header/header.css
@@ -3,6 +3,7 @@
   top: 0;
   left: 0;
   width: 100%;
+  height: 45px;
   z-index: 1000;
   display: flex;
   justify-content: space-between;
@@ -57,4 +58,9 @@
 .ant-menu-horizontal .ant-menu-item:hover {
   color: rgba(253, 253, 253, 0.639) !important;
   border-bottom-color: #808080 !important;
+}
+
+.header-dropdown {
+  position: relative;
+  right: 20px;
 }
--- a/src/main.jsx
+++ b/src/main.jsx
@@ -20,19 +20,11 @@ root.render(
           <Route index element={<LandingPage />} />
           <Route
             path="login"
-            element={
-              <AlreadyLoggedIn>
-                <LoginPage />
-              </AlreadyLoggedIn>
-            }
+            element={<AlreadyLoggedIn component={LoginPage} />}
           />
           <Route
             path="register"
-            element={
-              <AlreadyLoggedIn>
-                <RegisterPage />
-              </AlreadyLoggedIn>
-            }
+            element={<AlreadyLoggedIn component={RegisterPage} />}
           />
           <Route
             path="analysis"
--- a/src/main.jsx
+++ b/src/main.jsx
@@ -1,5 +1,6 @@
 import ReactDOM from "react-dom/client";
 import App from "./App.jsx";
+import { AuthProvider } from "./utils/authContext";
 
 import { BrowserRouter as Router, Routes, Route } from "react-router-dom";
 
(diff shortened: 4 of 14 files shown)


src/utils/protectedRoute.jsx
```diff
@@ -1,11 +1,18 @@
...
-  const { isLoggedIn } = useAuth();
+  const { isLoggedIn, loading } = useAuth();
 
+  //wait for confrimLogin to finish before rendering the component
+  if (loading) {
+    return <Spin size="large" />;
+  }
+
+  // 'replace' modifies the history stack so you can't return to the analysis page via the back button.
   return isLoggedIn ? <Component /> : <Navigate to="/login" replace />;
```
ErikAmerico: you can actually remove replace and you won't return to the analysis page while not logged in.

But apparently its a good safeguard to keep here for future components and situations.

ErikAmerico:
We should probably do something about navigating to the login or register page while you're already logged in. 

Maybe just redirect to the home page?

** I did this in the commit below

ErikAmerico:
the widening of the login/register forms doesn't really belong in this PR. 

But it was super quick, and Elle said that because it mentioned the word "login", that it was okay.

src/main.jsx
```diff
@@ -17,8 +18,22 @@ root.render(
...
+            path="register"
+            element={
+              <AuthRedirectRoute>
+                <RegisterPage />
+              </AuthRedirectRoute>
+            }
+          />
           <Route
             path="analysis"
             element={<ProtectedRoute component={AnalysisPage} />}
```
ErikAmerico: @rknizzle 

RegisterPage and LoginPage are wrapped in the AuthRedirectRoute component, AnalysisPage is passed to ProtectedRoute.

ProtectedRoute is waiting for confirmLogin to finish, I read that this method is preferred over wrapping the component, like as in where AuthRedirectRoute is only making a simple check for the logged in status

src/App.jsx
```diff
@@ -41,15 +40,13 @@ function App() {
   }, []);
 
   return (
-    <AuthProvider>
```
rknizzle, line 44: Why did you move AuthProvider to main instead of App? Is it because only things within the AuthProvider can use `useAuth` and now that is needed in main because of ProtectedRoute?

src/main.jsx
```diff
@@ -1,28 +1,30 @@
...
 
 import { BrowserRouter as Router, Routes, Route } from "react-router-dom";
 
 import LandingPage from "./components/LandingPage/LandingPage.jsx";
 import AnalysisPage from "./components/AnalysisPage/AnalysisPage.jsx";
 import LoginPage from "./components/loginRegister/LoginPage/LoginPage.jsx";
 import RegisterPage from "./components/loginRegister/RegisterPage/RegisterPage.jsx";
+import ProtectedRoute from "./utils/protectedRoute.jsx";
 
-export const RootComponent = () => {
```
rknizzle, line 11: Before your change here, the RootComponent was being `export`ed. I don't think anything is being exported anymore. Does that matter? Do you know if the export was doing anything before?

src/utils/protectedRoute.jsx
rknizzle: Ok lets update that comment then with what you know now

src/utils/protectedRoute.jsx
```diff
@@ -1,11 +1,18 @@
...
 
 //This component will be used to protect routes that require you to be signed in.
 const ProtectedRoute = ({ component: Component }) => {
-  const { isLoggedIn } = useAuth();
+  const { isLoggedIn, loading } = useAuth();
 
+  //wait for confrimLogin to finish before rendering the component
+  if (loading) {
+    return <Spin size="large" />;
+  }
```
rknizzle, line 13: This is a cool change. Just out of curiosity, did you ever run into a scenario where confirmLogin was still running while you were trying to navigate to the page?

rknizzle: Like have you actually been able to see this UI when running the app?

src/utils/authRedirectRoute.jsx
```diff
@@ -0,0 +1,27 @@
+import { useAuth } from "./useAuth";
```
rknizzle: Maybe change the name to something with like alreadyLoggedIn or something like that in it. I wasn't really sure what an authRedirect route would mean at first.

rknizzle (commented):
NIce! Good solution to this. I just have a few small comments and questions

rknizzle:
> the widening of the login/register forms doesn't really belong in this PR.
> 
> But it was super quick, and Elle said that because it mentioned the word "login", that it was okay.

lol

Yea thats totally cool. Feel free to make quick unrelated changes in PRs. Another good example would be a bug fix. If you find a bug unrelated to your current branch you can still just fix it and put it in your current PR. It doesn't need to be put into a whole separate branch and PR and stuff. Unless the bug fix would require a really big change then you might want to create a whole new branch/PR for it

src/utils/protectedRoute.jsx
rknizzle: @ErikAmerico 
github is kind of weird and my above comment was posted by itself as part of my review below so I just wanted to point out here that that comment is in response to your comment here ^^^^^^^^^

src/main.jsx
rknizzle: Can you explain your second sentence here again? I'm not fully following

ErikAmerico: Sorry @rknizzle 

I originally thought that only ProtectedRoutes was depending on the asynchronous confirmLogin function. And that AuthRedirectRoute was simply checking the state of isLoggedIn.

It was my understanding that if a component relied on an async function to render, then passing that component to ProtectedRoutes is the best option. 

If a component wasn't relying on an async function to render, you can just wrap it around the component and render it like I did with login and register.

But now that I am going over this again. 

Don't all of these technically rely on confirmLogin()? 

...

With the analysis page, I just know for a fact that it has to wait for confirmLogin to finish, because before I added the finally block to the confirmLogin function to have protectedroute wait for it to finish, I was getting all kinds of issues with timing and renders.

src/App.jsx
ErikAmerico: Yeah that was the initial reason for the switch, 

But then after looking into it more. You're actually supposed to have your AuthProvider at the root of the application.

So main is where it is supposed to be instead of app.jsx

src/utils/protectedRoute.jsx
ErikAmerico: Yeah I think if you're on the analysis page and just keep hitting the refresh browser button quickly, then you'll see it for a fraction of a second each time haha 

It could probably be centered on the page

src/main.jsx
ErikAmerico: Everything I just said could be completely wrong lol

But that's why I have them written differently

rknizzle: True yea I think they're kind of the same in that they both wait for confirmLogins result. They can probably both be written with the same pattern too whereas right now one of them takes in children and the other one takes in a component as a prop or something

src/main.jsx
ErikAmerico: The export wasn't doing anything before. 

This file is the entry point to the application where react connects to index.html, defines routes and now sets up AuthProvider.

I incorrectly exported RootComponent in the previous version. 

We live, We learn.

index.html
```diff
@@ -3,7 +3,10 @@
   <head>
     <meta charset="UTF-8" />
     <link rel="icon" type="image/svg+xml" href="assets/favicon.png" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1.0, user-scalable=no"
+    />
```
ErikAmerico, line 9: This was only an issue on touch screen devices, But it was a sloppy issue.

src/components/Header/header.css
```diff
@@ -3,6 +3,7 @@
   top: 0;
   left: 0;
   width: 100%;
+  height: 45px;
```
ErikAmerico, line 6: I set the height here because when the DropDown would render in the header, the header height would shrink, leaving a space between the header and the content.

 Setting the height of the header-container to 45px really causes the element to be 57px high due to the padding. It is now consistent across browsers and screen sizes.

 I'm not sure if we should leave headerHeight in app.jsx as a piece of state that can be updated depending on what we do in the future, or if we should just explicitly set the height like we did with FOOTER_HEIGHT.

But This works just fine for now.

ErikAmerico:
I feel like I'm taking advantage of your comment:

> Yea thats totally cool. Feel free to make quick unrelated changes in PRs.

I'm starting to get really off track in this PR. I'll close the laptop now

src/components/Header/header.css
rknizzle: I don't really have enough experience in this to give any input here so this is good if it works for now but yea just keep this in mind as you learn more and stuff and maybe we'll eventually change it to what you're thinking

src/utils/authRedirectRoute.jsx
rknizzle: I just changed it to AlreadyLoggedInRedirect

PR #57, opened 2024-04-30
Add border lines to model

rknizzle:
Add some darker border lines on the model which makes it easier to see the features of the model

Old:
old-viewer

New:
new-viewer


ErikAmerico (approved):
Sweet!

Getting better daily.

PR #58, opened 2024-05-01
View thickness visualization

ErikAmerico:
This works! It's not super pretty yet. 

________________________________

The only error I'm getting is 

IMG_8851

It only happens on the first time you click a threshold history item from the selection area. It doesn't effect the result and everything works. It might just be a react timing issue the resolves itself.  I'll keep looking into it

### DIFF
--- a/src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
+++ b/src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
@@ -5,29 +5,41 @@ import "./3dModelViewer.css";
 
 //TODO: Add loading spinner if a file is taking a second to load
 
-const ModelViewer = ({ fileFor3dModel }) => {
+const ModelViewer = ({ fileFor3dModel, fileNameForUpload }) => {
   const canvasRef = useRef(null);
   const [objectURL, setObjectURL] = useState(null);
+  const [fileType, setFileType] = useState(null);
 
   useEffect(() => {
     if (fileFor3dModel) {
       const url = URL.createObjectURL(fileFor3dModel);
       setObjectURL(url);
+      setFileType(fileNameForUpload.split(".").pop().toLowerCase());
 
       return () => {
         URL.revokeObjectURL(url);
       };
     }
+
+    // if i add fileNameForUpload in the dependency array, like react wants me to,
+    // everything breaks and i have to run docker compose down
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [fileFor3dModel]);
 
   useEffect(() => {
     if (!objectURL) return;
 
     // setupThreeScene returns a cleanup function
-    const cleanup = setupThreeScene(canvasRef.current, objectURL);
+    const cleanup = setupThreeScene(canvasRef.current, objectURL, fileType);
 
     // Call cleanup function when the component unmounts/before re-running the effect
     return () => cleanup();
+
+    // if i add fileType in the dependency array, like react wants me to,
+    // everything breaks and i have to run docker compose down
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [objectURL]);
 
   return (
@@ -39,6 +51,7 @@ const ModelViewer = ({ fileFor3dModel }) => {
 
 ModelViewer.propTypes = {
   fileFor3dModel: PropTypes.object,
+  fileNameForUpload: PropTypes.string,
 };
 
 export default ModelViewer;
--- a/src/components/AnalysisPage/3dModelViewer/setupThreeScene.js
+++ b/src/components/AnalysisPage/3dModelViewer/setupThreeScene.js
@@ -1,8 +1,9 @@
 import * as THREE from "three";
 import { STLLoader } from "three/examples/jsm/loaders/STLLoader.js";
+import { PLYLoader } from "three/examples/jsm/loaders/PLYLoader.js";
 import { OrbitControls } from "three/examples/jsm/controls/OrbitControls.js";
 
-const setupThreeScene = (canvas, objectURL) => {
+const setupThreeScene = (canvas, objectURL, fileType) => {
   const scene = new THREE.Scene();
   scene.background = new THREE.Color(0xffffff);
   const camera = new THREE.PerspectiveCamera(
@@ -46,22 +47,27 @@ const setupThreeScene = (canvas, objectURL) => {
 
   const disposables = [];
 
-  const loader = new STLLoader();
+  const loader = fileType === "stl" ? new STLLoader() : new PLYLoader();
+
   loader.load(objectURL, function (geometry) {
-    const material = new THREE.MeshPhongMaterial({
-      // wireframe: true,
+    const materialOptions = {
       specular: 0xffffff,
       shininess: 100,
-    });
-    disposables.push(geometry, material);
+      flatShading: true,
+    };
+
+    if (fileType === "ply") {
+      geometry.computeVertexNormals();
+      materialOptions.vertexColors = true;
+    }
+
+    const material = new THREE.MeshPhongMaterial(materialOptions);
 
     const mesh = new THREE.Mesh(geometry, material);
 
     geometry.center();
 
     scene.add(mesh);
-
-    disposables.push(mesh);
 
     // Adjusting the scale based on the model's bounding sphere
     geometry.computeBoundingSphere();
@@ -73,7 +79,7 @@ const setupThreeScene = (canvas, objectURL) => {
     const controls = new OrbitControls(camera, renderer.domElement);
     controls.addEventListener("change", () => renderer.render(scene, camera));
 
-    disposables.push(controls);
+    disposables.push(geometry, material, mesh, controls);
 
     const animate = () => {
       requestAnimationFrame(animate);
--- a/src/components/AnalysisPage/AnalysisPage.jsx
+++ b/src/components/AnalysisPage/AnalysisPage.jsx
@@ -16,6 +16,7 @@ const AnalysisPage = () => {
   //initialize as an array to use the spread operator in startButton.jsx
   const [listOfThicknessData, setListOfThicknessData] = useState([]);
   const [units, setUnits] = useState(null);
+  const [fileNameForUpload, setFileNameForUpload] = useState("");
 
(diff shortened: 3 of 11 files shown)


src/components/AnalysisPage/AnalysisPage.jsx
```diff
@@ -16,6 +16,7 @@ const AnalysisPage = () => {
   //initialize as an array to use the spread operator in startButton.jsx
   const [listOfThicknessData, setListOfThicknessData] = useState([]);
   const [units, setUnits] = useState(null);
+  const [fileNameForUpload, setFileNameForUpload] = useState("");
```
rknizzle, line 20: by 'ForUpload' are you meaning for the 3D viewer? The term upload makes it sound like the filename is for uploading something to the server

src/components/AnalysisPage/ThicknessAnalysisBox/StartButton/StartButton.jsx
```diff
@@ -17,10 +17,26 @@ const StartButton = ({
...
   const [showFailure, setShowFailure] = useState(false);
 
+  const downloadPlyFilePlaceIn3dViewer = (jobId) => {
+    fetch(`/api/v0/thickness/${jobId}/visual/ply`)
+      .then((res) => res.blob())
+      .then((blob) => {
+        // I can't yet locate a file name for the downloaded file.
+        // so I'm just setting it to "file.ply" for now.
+        // it works, and assuming if this function runs,
+        // the file should be a .ply file anyways, right?
```
rknizzle, line 39: We talked about this over text but yea the name of the file doesn't really matter here. Its just the file data thats important. 

Update or remove this comment if you don't already do it in a later commit

src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
```diff
@@ -5,29 +5,41 @@ import "./3dModelViewer.css";
...
 
       return () => {
         URL.revokeObjectURL(url);
       };
     }
+
+    // if i add fileNameForUpload in the dependency array, like react wants me to,
+    // everything breaks and i have to run docker compose down
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
rknizzle, line 27: I'll have to look into this. I'm curious what is making docker compose down be required again

src/components/AnalysisPage/3dModelViewer/setupThreeScene.js
```diff
@@ -46,23 +47,28 @@ const setupThreeScene = (canvas, objectURL) => {
 
   const disposables = [];
 
-  const loader = new STLLoader();
+  const loader = fileType === "stl" ? new STLLoader() : new PLYLoader();
```
rknizzle, line 50: This could be made more explicit. so like actually checking for a fileType of ply when choosing the PLYLoader rather than just defaulting to it if its not STL. And then if the fileType is neither stl or ply it could throw an error. This way if we later modify something and the fileType becomes weird it will be a lot easier to understand the issue. If the fileType becomes wrong then we'll see an error rather than it just defaulting to ply viewer

rknizzle (commented):
This feature is incredibly sick.

Heres a little review of the first part of the code. I'm probably going to merge this in though and then do a review of the rest of the code after the fact

src/components/AnalysisPage/AnalysisPage.jsx
ErikAmerico: Yeah uploading to the 3d viewer. I'll fix that in the next PR!

src/components/AnalysisPage/ThicknessAnalysisBox/StartButton/StartButton.jsx
ErikAmerico: I'll remove these in the next pr as well! 

considering this branch is about to be merged in

src/components/AnalysisPage/ThicknessAnalysisBox/indexedDBBlobStorage.js
```diff
@@ -0,0 +1,89 @@
...
+        } else {
+          console.log("No blob found for key:", key);
+          resolve(null); // or reject if you prefer to treat no data as an error
+        }
+      };
+      transaction.oncomplete = function () {
+        db.close();
+      };
+      transaction.onerror = function (event) {
+        console.error("Transaction failed:", event);
```
rknizzle, line 73: Should we reject here as well?

src/components/AnalysisPage/ThicknessAnalysisBox/indexedDBBlobStorage.js
```diff
@@ -0,0 +1,89 @@
...
+      };
+    });
+  } catch (error) {
+    console.error("Error during retrieval operation:", error);
+    throw error;
+  }
```
rknizzle, line 79: I don't think this try catch is really necessary unless you like this extra console log here. But I think the thrown error will probably be enough info.

Also I can't really remember how `return new Promise` works but I think that any reject() calls wont go into this catch block. But im not sure. So possibly only the lines above that will go into this catch block if they throw an error

src/components/AnalysisPage/ThicknessAnalysisBox/indexedDBBlobStorage.js
```diff
@@ -0,0 +1,89 @@
...
+      };
+      request.onsuccess = function () {
+        console.log("Blob stored successfully with key:", key);
+        resolve(request.result);
+      };
+      transaction.oncomplete = function () {
+        db.close();
+      };
+      transaction.onerror = function (event) {
+        console.error("Transaction failed:", event);
```
rknizzle, line 40: Same here. Should there also be a reject for this case?

src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
```diff
@@ -89,6 +90,21 @@ const ThickAnalysisBox = ({
...
 
+  async function loadPLYFromIndexedDB(partId, units, thresholdValue) {
+    try {
+      const blob = await retrieveBlob(partId, units, thresholdValue);
+      if (blob) {
+        setFileNameForUpload("file.ply");
+        const file = new File([blob], "file.ply", { type: blob.type });
+        setFileFor3dModel(file);
+      } else {
+        console.error("No PLY file found for this threshold.");
```
rknizzle: So maybe this is where the download from the server could go? We'd have to get the jobId for this job and then download with that job id

rknizzle (commented):
@ErikAmerico 
Nice! Heres my review of the second half of the code. Its looking good overall. We'll just have to solve the downloading the ply file from the server when its not in indexedDB

src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
rknizzle: I didn't get any server breaking errors like you descibe but I did start getting some errors in the console when I added those varibales into the dependency array. useEffect is something I'm still not super good with. I'll learn about it more over time. I'm guessing the fact that we are needing to hide the linter could be a sign that we should be using better patterns to not have to disable it but I don't know if thats true currently

src/components/AnalysisPage/3dModelViewer/setupThreeScene.js
ErikAmerico: @rknizzle 

Yes for sure. 

The only issue is when the file type isn't supported, the whole application breaks. 

I'm looking into wrapping the modelViewer component in an error-boundary component to display some sort of UI instead of everything falling apart.

But we probably don't even need that error-boundary component's UI because we would never deploy code that sends a bad file type to the 3dModelViewer... We'll just see the error and fix it.

Right?

src/components/AnalysisPage/ThicknessAnalysisBox/indexedDBBlobStorage.js
ErikAmerico: yes to this, and to the similar comment below!

src/components/AnalysisPage/ThicknessAnalysisBox/indexedDBBlobStorage.js
ErikAmerico: unbelievable, you're right again. 

But I'm reading that we don't even need to throw the error because it is redundant after rejecting the promise. 
Does that sound accurate?

So i'm just going to remove the try-catch from both storeBlob and retrieveBlob.

rknizzle: Its not just redundant for the rejected promise, the rejection never actually goes into the catch block.

For example if you run this code:
```js
function testPromiseBehavior() {

	try {
		return new Promise((resolve, reject) => {
			console.log('1')
			return reject('bad thing')
		})
	} catch(err) {
		console.log('2')
	}
}

testPromiseBehavior()
```

It only logs 1 and doesn't log 2

BUT for the other stuff above the 'return new Promise' statement; if anything in there throws an error then it will still go into the catch block.
See this example:
```js
async function testPromiseBehavior() {

	try {

		doOtherThing()

		return new Promise((resolve, reject) => {
			console.log('1')
			return reject('bad thing')
		})
	} catch(err) {
		console.log('2')
	}
}

testPromiseBehavior()

function doOtherThing() {
	throw new Error('a bad thing from doOtherThing')
}

```
This will log 2.


And also if you don't already know; a catch block with just rethrowing is totally redundant. It doesn't do anything.

```js
try {
  doThingThatMightError()
} catch(err) {
  // this whole catch statement does nothing and the try catch should just be removed
  throw err
}
```

@ErikAmerico 
Let me know if anything here doesn't make sense.

But so yea in this case you can porbably just remove the try catches because all theyre doing is adding a log when one of the things above 'return new Promise' throws an error

src/components/AnalysisPage/3dModelViewer/setupThreeScene.js
rknizzle: Yea I think you're right here. It probably wont happen that an unsupported file type would be passed to here. It'll just be something that we need to fix if it ever does happen

PR #63, opened 2024-05-02
Download ply file from server when not in indexed db

ErikAmerico:
Get the ply file from the server if its not stored in indexedDB.

I'll make another branch tomorrow to address the rest of your comments in the previous pull request!


src/components/AnalysisPage/ThicknessAnalysisBox/downloadPlyToViewer.js
```diff
@@ -0,0 +1,23 @@
...
+    .then((res) => res.blob())
+    .then((blob) => {
+      setFileNameForUpload("file.ply");
+      const file = new File([blob], "file.ply", { type: blob.type });
+      setFileFor3dModel(file);
+      storeBlob(partId, units, thresholdValue, blob);
+    })
+    .catch((error) => {
+      console.error("Error downloading the PLY file:", error);
+    });
```
rknizzle: I'd recommend either throwing the error again after this log statement or just getting rid of the whole .catch here. This is doing something called 'error swallowing' where you are catching errors and then you're not passing the error up to whatever called this function. So I think that whatever function calls this function will think that it succeeded because it doesn't throw an error.

I believe that you'll also need to add a return to the fetch:
```js
return fetch(`/api/v0/thickness/${jobId}/visual/ply`)
....
```

So that the response of the fetch call gets returned up to the caller

rknizzle: Also I have a tip for making code changes more easily reviewable.

If you are going to move a whole function to another file or to another place in the file; just do the movement and then commit it and say that the function got moved. And then make any changes after that and put the changes into another commit. That way I can see the diff of the actual changes to the function. When you combine the changes with a movement, the entire function becomes a big green addition and so it doesn't show specifically what got changed in the function anymore.

I'm guessing this'll take a few mentions before it becomes a habit but I just wanted to start floating this idea

src/components/AnalysisPage/ThicknessAnalysisBox/StartButton/StartButton.jsx
```diff
@@ -5,8 +5,8 @@ import "./startButton.css";
 import PropTypes from "prop-types";
```
rknizzle, line 5: I'm looking at commit 4a611ab4fa7379ca79b5db0195f87a0aa48da9f4:

I think that DecimalInput.jsx can stay in the regular directory instead of utils since its a React component. Its just a preference thing but utils to me feel more like just small functions that do a small specific task used in multiple places whereas DecimalInput is just a component. The other functions feel a bit more like utils if thats where you want to put them. I'm indifferent to putting them into utils or just leaving them in the ThicknessAnalysisBox directory

src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
```diff
@@ -57,21 +57,48 @@ const ThickAnalysisBox = ({
...
+          downloadPlyFilePlaceIn3dViewer(
+            data.id,
+            data.part_id,
+            data.units,
+            data.threshold
+          );
+        }
       }
     } catch (error) {
       console.error("Failed to load PLY from IndexedDB:", error);
```
rknizzle, line 93: We may want to add an error popup in the UI if the download fails rather than just a log

rknizzle (commented):
Cool good fix! Just a couple suggestions here nothing big

rknizzle:
@ErikAmerico I also pushed a couple refactoring commits too just while I was going through and brushing up on the thickness analysis box code again. You can review it if you want

src/components/AnalysisPage/ThicknessAnalysisBox/downloadPlyToViewer.js
ErikAmerico: Gotchya! 

Yeah that makes too much sense

src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
ErikAmerico: We took care of this in downloadPlyToViewer.js in a new commit below!

src/utils/startPageAtTop.js
```diff
@@ -0,0 +1,12 @@
+import { useEffect } from "react";
+import { useLocation } from "react-router-dom";
+
```
ErikAmerico, line 3: Just as an example:
 
Before this commit if you were halfway down the landing page and clicked on 'Analyze Parts'
You would then be brought to the analysis page, but positioned in the middle of the page instead of at the top.

This really was only an issue on smaller screens, But it was a sloppy issue.

rknizzle:
@ErikAmerico 
Nice! The feedback here is all taken care of. I'm going to merge this in.

Reminder that theres also a couple review comments left in https://github.com/rknizzle/propermesh-view/pull/58 when you have a chance to go through it. No rush

PR #65, opened 2024-05-04
add toggle to switch between ply file and original file in 3d viewer

ErikAmerico:
Added a toggle button to switch between viewing the ply file and original file.

### DIFF
--- a/src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
+++ b/src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
@@ -2,13 +2,28 @@ import { useEffect, useRef, useState } from "react";
 import setupThreeScene from "./setupThreeScene";
 import PropTypes from "prop-types";
 import "./3dModelViewer.css";
+import { Switch } from "antd";
 
 //TODO: Add loading spinner if a file is taking a second to load
 
-const ModelViewer = ({ fileFor3dModel, fileNameFor3dModel }) => {
+const ModelViewer = ({
+  fileFor3dModel,
+  fileNameFor3dModel,
+  originalFileFor3dModel,
+  originalFileNameFor3dModel,
+  showToggle,
+  setShowToggle,
+}) => {
   const canvasRef = useRef(null);
   const [objectURL, setObjectURL] = useState(null);
   const [fileType, setFileType] = useState(null);
+
+  useEffect(() => {
+    if (fileFor3dModel === originalFileFor3dModel) {
+      setShowToggle(false);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [fileFor3dModel, originalFileFor3dModel]);
 
   useEffect(() => {
     if (fileFor3dModel) {
@@ -42,9 +57,36 @@ const ModelViewer = ({ fileFor3dModel, fileNameFor3dModel }) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [objectURL]);
 
+  const handleToggleChange = (checked) => {
+    if (checked) {
+      const url = URL.createObjectURL(fileFor3dModel);
+      setObjectURL(url);
+      setFileType(fileNameFor3dModel.split(".").pop().toLowerCase());
+
+      return () => {
+        URL.revokeObjectURL(url);
+      };
+    }
+
+    if (!checked) {
+      const url = URL.createObjectURL(originalFileFor3dModel);
+      setObjectURL(url);
+      setFileType(originalFileNameFor3dModel.split(".").pop().toLowerCase());
+
+      return () => {
+        URL.revokeObjectURL(url);
+      };
+    }
+  };
+
   return (
     <div id="model-viewer-container">
       <canvas ref={canvasRef} id="model-canvas" />
+      {showToggle && (
+        <div className="model-viewer-toggle">
+          <Switch defaultChecked onChange={handleToggleChange} />
+        </div>
+      )}
     </div>
   );
 };
@@ -52,6 +94,10 @@ const ModelViewer = ({ fileFor3dModel, fileNameFor3dModel }) => {
 ModelViewer.propTypes = {
   fileFor3dModel: PropTypes.object,
   fileNameFor3dModel: PropTypes.string,
+  originalFileFor3dModel: PropTypes.object,
+  originalFileNameFor3dModel: PropTypes.string,
+  showToggle: PropTypes.bool,
+  setShowToggle: PropTypes.func,
 };
 
 export default ModelViewer;
--- a/src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
+++ b/src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
@@ -17,10 +17,8 @@ const ModelViewer = ({
   const canvasRef = useRef(null);
   const sceneRef = useRef(null);
   const [previousFileFor3dModel, setPreviousFileFor3dModel] = useState(null);
-  const [
-    previousOriginalFileNameFor3dModel,
-    setPreviousOriginalFileNameFor3dModel,
-  ] = useState(null);
+  const [previousOriginalFileFor3dModel, setPreviousOriginalFileFor3dModel] =
+    useState(null);
 
   useEffect(() => {
     sceneRef.current = setupThreeScene(canvasRef.current);
@@ -29,12 +27,11 @@ const ModelViewer = ({
   useEffect(() => {
     const isNewFileFor3dModel =
       fileFor3dModel && fileFor3dModel !== previousFileFor3dModel;
-    const isNewOriginalFileNameFor3dModel =
+    const isNewOriginalFileFor3dModel =
       originalFileFor3dModel &&
-      originalFileFor3dModel !== previousOriginalFileNameFor3dModel;
+      originalFileFor3dModel !== previousOriginalFileFor3dModel;
 
-    const viewingNewPart =
-      isNewFileFor3dModel && isNewOriginalFileNameFor3dModel;
+    const viewingNewPart = isNewFileFor3dModel && isNewOriginalFileFor3dModel;
 
     if (fileFor3dModel) {
       const fileType = fileNameFor3dModel.split(".").pop().toLowerCase();
@@ -43,7 +40,7 @@ const ModelViewer = ({
       URL.revokeObjectURL(objectURL);
       sceneRef.current.clearModel();
       setPreviousFileFor3dModel(fileFor3dModel);
-      setPreviousOriginalFileNameFor3dModel(originalFileFor3dModel);
+      setPreviousOriginalFileFor3dModel(originalFileFor3dModel);
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [fileFor3dModel]);
--- a/src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
+++ b/src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef } from "react";
+import { useEffect, useRef, useState } from "react";
 import setupThreeScene from "./setupThreeScene";
 import PropTypes from "prop-types";
 import "./3dModelViewer.css";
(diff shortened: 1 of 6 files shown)


src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
```diff
@@ -76,6 +77,7 @@ const ThickAnalysisBox = ({
         setFileNameFor3dModel("file.ply");
         const file = new File([blob], "file.ply", { type: blob.type });
         setFileFor3dModel(file);
+        setShowToggle(true);
       } else {
         if (listOfThicknessData.find((data) => data.id === data.id)) {
```
ErikAmerico: if downloadPlyFilePlaceIn3dViewer fails, will setShowToggle(true) still run? or will this whole if statement just cancel?

I wasn't able to pass setShowToggle to downloadPlyFile to handle this in that function

ErikAmerico:
Oh and we don't need to add this toggle button, I just know you mentioned having one. 

Up to you!

src/components/AnalysisPage/ThicknessAnalysisBox/ThickAnalysisBox.jsx
rknizzle: If downloadPlyFilePlaceIn3dViewer throws an error then setShowToggle(true) will not run. Throwing an error stops the code executing. When an error is thrown it 'bubbles up' until it goes up into a try in which case it will stop bubbling up and go into the catch block.

Heres an example of bubbling up:

```js
function top() {
  try {
    doThing()
  } catch(err) {
    console.log('catching the error, maybe there would be a UI popup or something')
  }
}

function lower() {
  lowest()
  console.log('2')
  otherThing()
}

function lowest() {
  console.log('1')
  throw new Error('a bad thing happened')
}


top()

```

This program logs '1' and then 'catching the error, maybe there would be a UI popup or something'.

It does not log '2'.

So  this shows an error bubbling up. As functions call other functions it creates new lower layers and so when one of those lower layers throws an error, it goes back up to the higher layers either until it goes into a try in which case it will stop bubbling up and go into the catch block or iif it never goes into a try block itll just bubble up until it gets to the top which would be probably the React component

rknizzle:
@ErikAmerico I did want a toggle but now with how you made the UI easily switch back to grey when you delete a value in the threshold input I don't t hink its as necessary but it might still be good. What do you think about the toggle UI?

But I do think that I definitely don't want to release the toggle feature until we figure out how to change the color of the model without reloading the whole scene so it doesn't cause the black screen flash and the resetting of the model position. If we add the toggle I'd like it to just change the color and not reset the position of the model

ErikAmerico:
I don't think the toggle is completely useless.

If someone has a 1 threshold value result saved and they are currently viewing a 1.7. When they wanted to switch back to the gray model, 

they wild have to delete the .7, which would briefly show the ply file for the 1 value then they can delete the 1 to get the gray model.

and I personally felt trapped when viewing the ply model. Now I feel like I'm free to look at either 😂

ErikAmerico:
I'll start getting into switching the color without resetting the scene! And keeping the same position.

src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
```diff
@@ -2,14 +2,29 @@ import { useEffect, useRef, useState } from "react";
...
+  setShowToggle,
+}) => {
   const canvasRef = useRef(null);
   const [objectURL, setObjectURL] = useState(null);
   const [fileType, setFileType] = useState(null);
 
+  useEffect(() => {
+    if (fileFor3dModel === originalFileFor3dModel) {
+      setShowToggle(false);
+    }
```
rknizzle: Can showToggle and setShowToggle be totally managed by the ModelViewer component? Does it need to be state in the Analysis page and passed around to a bunch of other components?

src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
```diff
@@ -15,67 +15,48 @@ const ModelViewer = ({
   setShowToggle,
 }) => {
   const canvasRef = useRef(null);
-  const [objectURL, setObjectURL] = useState(null);
-  const [fileType, setFileType] = useState(null);
+  const sceneRef = useRef(null);
```
rknizzle, line 15: Why is this a 'Ref' and why does it use the property current? Why isn't it just like modelViewerSceneHandler or something? Is there some reason why thats necessary? I'm not super familiar with what refs are in React

src/components/AnalysisPage/3dModelViewer/setupThreeScene.js
```diff
@@ -138,11 +148,15 @@ const setupThreeScene = (canvas, objectURL, fileType) => {
...
-    renderer.dispose();
-
-    scene.clear();
+  const animate = () => {
+    requestAnimationFrame(animate);
+    renderer.render(scene, camera);
   };
+  animate();
+
+  return { loadModel, clearModel };
```
rknizzle: Really clever solution here returning an object containing functions to modify the scene from the React component

src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
```diff
@@ -15,67 +15,48 @@ const ModelViewer = ({
...
-      setFileType(fileNameFor3dModel.split(".").pop().toLowerCase());
-
-      return () => {
-        URL.revokeObjectURL(url);
-      };
+      const fileType = fileNameFor3dModel.split(".").pop().toLowerCase();
+      const objectURL = URL.createObjectURL(fileFor3dModel);
+      sceneRef.current.loadModel(objectURL, fileType);
+      URL.revokeObjectURL(objectURL);
+      sceneRef.current.clearModel();
```
rknizzle: Wait why are you loading and then clearing right after?

src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
```diff
@@ -17,10 +17,8 @@ const ModelViewer = ({
   const canvasRef = useRef(null);
   const sceneRef = useRef(null);
   const [previousFileFor3dModel, setPreviousFileFor3dModel] = useState(null);
-  const [
-    previousOriginalFileNameFor3dModel,
-    setPreviousOriginalFileNameFor3dModel,
-  ] = useState(null);
+  const [previousOriginalFileFor3dModel, setPreviousOriginalFileFor3dModel] =
```
rknizzle: Just a note on commit message for cbdfbceb7b2f542f73557f64b600f202f9bbc00a:
You renamed 2 variables in the commit but only mentioned one of them in the commit message. So just be careful of that cause it can cause a little confusion. You could have either split this into 2 commits or just mentioned both renames in the commit.

I use the command `git diff --staged` very often (I created a terminal shortcut `gds` for it). It shows all the changes that you have added and are going to get commited when you run `git commit`. This can be really useful for checking whats about to be committed.

Also commits contain sort of 2 parts; a title and a message. The title is just a one line thing and then the message can be as long as you want it to be. So you could have done like:

```
Refactor: Rename variables

previousOriginalFileNameFor3dModel -> previousOriginalFileFor3dModel
&
isNewOriginalFileNameFor3dModel -> isNewOriginalFileFor3dModel
```

Where the top line is the title and the rest is the message. 

The way that you commit is just kind of only a title that extends to the next line if it gets too long.

Heres an example of a commit that has a title and a message: https://github.com/rknizzle/propermesh/commit/77ae192b77760a33c3d7f64f1fd9563f4ba844d8

The title vs message thing isn't a big deal though so you don't need to worry about changing the way you commit right now. I just wanted to introduce you to the concept

src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
```diff
@@ -16,18 +16,34 @@ const ModelViewer = ({
...
 
   useEffect(() => {
+    const isNewFileFor3dModel =
+      fileFor3dModel && fileFor3dModel !== previousFileFor3dModel;
+    const isNewOriginalFileNameFor3dModel =
+      originalFileFor3dModel &&
+      originalFileFor3dModel !== previousOriginalFileNameFor3dModel;
+
+    const viewingNewPart =
+      isNewFileFor3dModel && isNewOriginalFileNameFor3dModel;
```
rknizzle: THis is starting to get a little confusing with all the current vs original vs previous stuff for everything. Im not quite sure how to make it better though. Maybe just refactoring to make some small well named functions or something.

Or maybe theres a way to split it apart so that instead of all the loading logic being here we could have like a function to load in a new original file and then a separate function to load the ply visualization file into the viewer.

Not sure just thinking out loud, we can keep an eye on this as we continue developing the viewer

src/components/AnalysisPage/3dModelViewer/setupThreeScene.js
```diff
@@ -47,86 +47,96 @@ const setupThreeScene = (canvas, objectURL, fileType) => {
...
+  const loadModel = (objectURL, fileType) => {
+    console.log("Loading model");
+    // clearModel();
+    let loader;
+    if (fileType === "stl") {
+      loader = new STLLoader();
+    } else if (fileType === "ply") {
+      loader = new PLYLoader();
+    } else {
+      throw new Error(`Unsupported file type`);
```
rknizzle: The toggle works a lot better now but I think theres one more optimization we can make when switching between the grey model and the thickness visualization model.  Once we load in a ply file, instead of calling the loader every time to load back in the original file to get the grey colored view, we can just change the color of the model using threejs materials. So instead of loading in the original part file every time to get the grey view we can just change the material on the mesh to show grey.

This will solve the little pause between switching out the models too. Currently if you load in a really big file it can take 4-5 seconds to switch between the views when you press the toggle because it needs to load the whole part into the loader each time.

You've already mentioned wanting to fix the little flash of the part when switching the toggle so I think this might be a good way to go about it

rknizzle (commented):
Nice solution! I left a couple little changes and some questions here and then I'll merge this in when you address those. I also gave my idea for an optimization to fix the pause between showing the model when you toggle the switch. This works well how it is though so I'm also down to merge this in before you fully fix the pause between models. As soon as you answer my couple little questions and possibly make some updates for them I'm good to merge this in

src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
ErikAmerico: yes it can, and super easily. Thanks dude!

src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
ErikAmerico: I'm also not an expert in react Refs.

But here its a reference to the setUpThreeScene function call.

To the best of my knowledge

Using ref lets us load and clear new models into the scene without having to re-setup the scene every time a new model loads or clears. Where if we were using a modelViewerSceneHandler state, it would trigger a re-render on each change and we would be calling setUpThreeScene over and over again (like we were before i believe, causing that black flash).

.current is where the reference is located in sceneRef. Here is the output of `console.log(sceneRef)` 

image

src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
ErikAmerico: hmmm really only because it just works this way. If i remove `sceneRef.current.clearModel()`, then the next time i load a model. the new model will be overlapping the previous model. 

But I think whats actually happening is that when this code runs, `sceneRef.current.clearModel()` doesn't actually have anything to clear because the model hasn't been added to sceneRef.current yet. Then the next time the useEffect runs, the new model will be in the process of loading while `sceneRef.current.clearModel()` gets hit and clears the previously loaded model.

src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
```diff
@@ -49,6 +48,9 @@ const ModelViewer = ({
     if (fileFor3dModel === originalFileFor3dModel) {
       setShowToggle(false);
     }
+    if (fileFor3dModel !== originalFileFor3dModel) {
```
rknizzle: I think just using an else here would be simpler and clearer

src/components/AnalysisPage/3dModelViewer/setupThreeScene.js
```diff
@@ -81,6 +82,15 @@ const setupThreeScene = (canvas) => {
 
       const mesh = new THREE.Mesh(geometry, material);
 
+      // this if statement has to be seperate from similar statement
+      // on line 76 because the materials have to be defined before the mesh
```
rknizzle: Try to avoid specifying particular line numbers in code because if you add a new feature in your next PR that pushes the code down, the correct code won't line up with this comment anymore. Try to just use descriptions instead

src/components/AnalysisPage/3dModelViewer/setupThreeScene.js
```diff
@@ -128,6 +138,16 @@ const setupThreeScene = (canvas) => {
     });
   };
 
+  const toggleMesh = (fileType) => {
```
rknizzle: I think this function could use a bit more detailed of a name to include what its toggling between. Just so in the future if we have like 10 more functions to modify the 3d model scene this will be more clear what its doing

rknizzle (changes requested):
The switch toggle between views is working really really good now! And yea you already mentioned the couple things you need to finish up now in your commit messages.
- removing state no longer used
- getting rid of the pause/flash when you edit the threshold value

Nice detailed commit messages now too!

src/components/AnalysisPage/3dModelViewer/3dModelViewer.jsx
ErikAmerico: In one of my latest commits, I was able to completely remove all of this! 

no more confusion

src/components/AnalysisPage/3dModelViewer/setupThreeScene.js
ErikAmerico: This should be good now too!

ErikAmerico:
Also when adjusting the threshold value, The gray mesh gets displayed instead of reloading the original file. 

There is still the issue of when you type in a value that already has results, It loads in the ply file and you get that black flash. 

I know we talked about it probably not getting fixed. But I feel like after working tonight, I will be able to figure it out!

rknizzle:
Nice! Looks good merging this in

PR #70, opened 2024-05-22
Model viewer fixes

ErikAmerico:
I would probably think that we should merge this PR even though its only 2 commits, right?

That units selector oversight shouldn't be on the internet.

### DIFF
--- a/src/components/AnalysisPage/AnalysisPage.jsx
+++ b/src/components/AnalysisPage/AnalysisPage.jsx
@@ -104,7 +104,15 @@ const AnalysisPage = () => {
         </Col>
         <Col xs={{ span: 24, order: 3 }} lg={{ span: 24, order: 5 }}>
           <div id="unit-selector-parent-small">
-            <UnitsSelector partId={partId} units={units} setUnits={setUnits} />
+            <UnitsSelector
+              partId={partId}
+              units={units}
+              setUnits={setUnits}
+              setFileFor3dModel={setFileFor3dModel}
+              setFileNameFor3dModel={setFileNameFor3dModel}
+              originalFileFor3dModel={originalFileFor3dModel}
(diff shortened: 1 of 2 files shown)


src/components/AnalysisPage/AnalysisPage.jsx
```diff
@@ -104,7 +104,15 @@ const AnalysisPage = () => {
         </Col>
         <Col xs={{ span: 24, order: 3 }} lg={{ span: 24, order: 5 }}>
           <div id="unit-selector-parent-small">
-            <UnitsSelector partId={partId} units={units} setUnits={setUnits} />
+            <UnitsSelector
```
rknizzle, line 107: Do you think theres a better way to handle modifying a component for different screen size than just specifying it twice so that stuff like this doesn't happen? Or does this seem to be the proper way to handle this scenario?

rknizzle (commented):
Good fixes

rknizzle:
@ErikAmerico Yes the fewer commits, the better in a PR.

 I haven't actually tested this code but I'll trust you that it works. Going to merge this in and deploy it tonight or tomorrow morning

src/components/AnalysisPage/AnalysisPage.jsx
ErikAmerico: I'm not sure if someone who was a front end expert would consider it proper or not. 

But in our scenario, It is proper if you ask me. The large screen UnitsSelector is inside of a column with the geometry analysis box and thickness anaylsis box, which is how we wanted the analysis page to be laid out. 

Without having a seperate small screen UnitsSelector seperate from the column with the geo-box and thickness-box. I wouldn't have been able to put it on the top of the page when viewing from a cell phone.

